PR #2, opened 2025-07-10
feat: support registering dev plugins, impl blobstore

brooksmtownsend:
This PR implements the ability to register plugins with a hook type `DevRegister`, which registers the exports and resources of that component for use with `wash dev`.

The plugin and example in this PR show how this works, with the plugin `blobstore-filesystem` exporting `wasi:blobstore` and importing `wasi:filesystem`, implementing the blobstore interface in terms of the FS bindings. After registering that plugin, any component that uses `wasi:blobstore` can work with `wash dev` since its requirements are being linked at runtime. Put another way, `wash dev` is now pluggable for ANY custom interface that can be implemented as a plugin (which is granted the 0.2 world minus HTTP right now.)

There's pretty particular management of resources and stores to accomplish this, including making sure to instantiate all new plugins for every single instance, lifting and lowering resource types to ensure compatibility, and skipping the polyfilling of host resources like `input-stream` and `output-stream`. There are likely more edge cases to find here.

## To test
1. Install `wash` from this branch
```
wash build ./plugins/blobstore-filesystem
wash plugin install ./plugins/blobstore-filesystem/target/wasm32-wasip2/release/blobstore_filesystem.wasm
wash dev ./examples/http-blobstore/
```
3. `curl localhost:8000 --data 'hi there!'` <-- streams the data that you sent to the filesystem and back


brooksmtownsend:
Note that the windows tests failing is because of something about `wash doctor` and resolving the npm tool on windows.

PR #37, opened 2025-08-15
dev(watch): canonicalize project root & add artifact-aware ignores

### LINKED ISSUE #20: Ignore artifact directories in file watcher
## Description

The development file watcher currently triggers rebuilds when files in artifact/build directories change. This can cause infinite rebuild loops since the build process itself generates files that trigger the watcher.

## Location

`crates/wash/src/cli/dev.rs:254` in the file watcher logic

## Current Behavior

```rust
// TODO(GFI): ensure we ignore the artifact dirs
// Ensure that paths that take place in ignored directories don't trigger a reload
// This is primarily here to avoid recursively triggering reloads for files that are
// generated by the build process.
```

The file watcher monitoring logic has commented-out code to prevent artifact directories from triggering rebuilds.

## Problem

Without proper ignore patterns:
- Build artifacts trigger unnecessary rebuilds
- Can cause infinite rebuild loops
- Wastes CPU and disk I/O resources
- Degrades developer experience with constant rebuilds

## Proposed Solution

Implement ignore patterns for common artifact directories:

```rust
let should_ignore = paths.iter().any(|path| {
    let relative_path = path.strip_prefix(&project_dir).unwrap_or(path);
    
    // Common build/artifact directories to ignore
    relative_path.components().any(|component| {
        matches!(component.as_os_str().to_str(), Some(
            "target" |           // Rust builds
            "build" |            // General builds  
            "dist" |             // Distribution builds
            "node_modules" |     // Node.js dependencies
            ".wash" |            // Wash artifacts
            ".git"               // Git metadata
        ))
    })
});

if should_ignore {
    return; // Skip triggering rebuild
}
```

## Implementation Notes

- Should be configurable via project config
- Consider using `.gitignore`-style patterns
- May need project-type-specific defaults (Rust vs TypeScript vs Go)

## Benefits

- **Prevents infinite rebuilds**: Build outputs won't trigger new builds
- **Better performance**: Reduces unnecessary file system monitoring
- **Cleaner logs**: Fewer spurious rebuild messages

navahas:
- Normalize project root once for stable path comparisons
- Build ignore set from project type + artifact parent dir
- Keep pause/reload semantics; filter before enqueue
- Add tests (rust/ts/mixed, artifact parent, rel vs abs, symlink, nested)

Closes #20

## Feature or Problem
Reduce self-triggered rebuilds in `wash dev` by ignoring generated outputs and comparing paths consistently. Canonicalizing the project root and adding project-aware ignore rules prevents noisy loops from `target/`, `dist/`, `node_modules/`, etc., and from the build artifact’s parent directory.

## Related Issues
- Closes #20

## Release Information
Included in the next release (no config or API changes).

## Consumer Impact
No breaking changes. Fewer spurious rebuilds; behavior is more predictable across platforms. Users may notice that edits inside build/output dirs no longer trigger rebuilds (intended).

### Unit Test(s)
Added/updated tests to cover:
- Rust, TypeScript, Mixed contexts
- Artifact parent directory ignoring
- Relative vs absolute path consistency
- Symlink handling
- Nested ignored directories

### Acceptance or Integration
No changes required.

### Manual Verification
Using the patched binary, scaffold a tiny component and start the dev loop.

```bash
# Set your workspace dir (no spaces around =)
DIR="$HOME/dev/wasmCloud"

# Copy the patched wash binary next to your projects
cp "$DIR/wash/target/debug/wash" "$DIR/wash-fix"

# Scaffold a minimal Rust HTTP component (uses the local binary)
cd "$DIR"
./wash-fix new http-hello-world --template-name hello-world-rust
```
#### Terminal A
In the component folder
```
cd "$DIR/http-hello-world"
RUST_LOG=trace,wash=trace ../wash-fix dev --address 127.0.0.1:8000
```
#### Terminal B
same folder

Real source change
- Terminal A: SHOULD rebuild
```
echo "// change $(date)" >> src/lib.rs
```

Ignored dirs 
- Terminal A: SHOULD NOT rebuild
```
mkdir -p target/tmp dist/tmp node_modules/tmp .wash/tmp .wasmbuild/tmp
touch target/tmp/a.txt dist/tmp/b.txt node_modules/tmp/c.txt .wash/tmp/d.txt .wasmbuild/tmp/e.txt
```

Artifact parent dir
- Terminal A: SHOULD NOT rebuild
```
ART="$(ls target/wasm32-wasi*/debug/*.wasm | head -n1)"
touch "$ART" "$(dirname "$ART")/probe.txt"
```

Similar name, not ignored
- Terminal A: SHOULD rebuild
```
mkdir -p targetx
touch targetx/should-trigger.txt
```

### DIFF
--- a/crates/wash/src/cli/dev.rs
+++ b/crates/wash/src/cli/dev.rs
@@ -56,39 +56,13 @@ use crate::{
 /// * `path` - The file path to check
 /// * `canonical_project_root` - The canonicalized project root directory
 /// * `ignore_paths` - Set of canonicalized paths that should be ignored
-fn is_ignored(path: &Path, canonical_project_root: &Path, ignore_paths: &HashSet<PathBuf>) -> bool {
-    // Try to get the canonical path, fall back to the original if canonicalization fails
+fn is_ignored(
+    path: &Path,
+    _canonical_project_root: &Path,
+    ignore_paths: &HashSet<PathBuf>,
+) -> bool {
     let canonical_path = path.canonicalize().unwrap_or_else(|_| path.to_path_buf());
-
-    // Check if the canonical path starts with any of the ignore paths
-    for ignore_path in ignore_paths {
-        if canonical_path.starts_with(ignore_path) {
-            return true;
-        }
-    }
-
-    // Also check relative paths from canonical project root
-    if let Ok(relative_path) = canonical_path.strip_prefix(canonical_project_root) {
-        // Check if any component of the relative path matches common build directories
-        for component in relative_path.components() {
-            if let Some(component_str) = component.as_os_str().to_str() {
-                match component_str {
-                    "target" |           // Rust builds
-                    "build" |            // General builds  
-                    "dist" |             // Distribution builds
-                    "node_modules" |     // Node.js dependencies
-                    ".wash" |            // Wash artifacts
-                    ".git" |             // Git metadata
-                    ".wasmbuild" => {     // WebAssembly build artifacts
-                        return true;
-                    }
-                    _ => {}
-                }
-            }
-        }
-    }
-
-    false
+    ignore_paths.iter().any(|p| canonical_path.starts_with(p))
 }
 
 /// Build a set of paths that should be ignored during file watching
@@ -103,27 +77,17 @@ fn is_ignored(path: &Path, canonical_project_root: &Path, ignore_paths: &HashSet
 /// A set of canonicalized paths that should be ignored during file watching
 fn build_ignore_set(
     canonical_project_root: &Path,
-    artifact_path: Option<&Path>,
     project_context: &ProjectContext,
 ) -> HashSet<PathBuf> {
     let mut ignore_paths = HashSet::new();
 
-    // Add the artifact path's parent directory if provided
-    if let Some(artifact) = artifact_path {
-        if let Some(artifact_dir) = artifact.parent() {
-            if let Ok(canonical) = artifact_dir.canonicalize() {
-                ignore_paths.insert(canonical);
-            }
-        }
-    }
-
     // Common directories for all project types
-    let mut dirs_to_ignore = vec![".git", ".wash"];
+    let mut dirs_to_ignore = vec![".git"];
 
     // Add project-type-specific ignore patterns
     match project_context {
         ProjectContext::Rust { .. } => {
-            dirs_to_ignore.extend_from_slice(&["target", ".wasmbuild"]);
+            dirs_to_ignore.extend_from_slice(&["target"]);
         }
         ProjectContext::TypeScript { .. } => {
             dirs_to_ignore.extend_from_slice(&["node_modules", "dist", "build", ".next", ".nuxt"]);
@@ -134,7 +98,7 @@ fn build_ignore_set(
         ProjectContext::Mixed { detected_types } => {
             // Add ignore patterns for all detected project types
             if detected_types.iter().any(|t| t == "Rust") {
-                dirs_to_ignore.extend_from_slice(&["target", ".wasmbuild"]);
+                dirs_to_ignore.extend_from_slice(&["target"]);
             }
             if detected_types
                 .iter()
@@ -159,7 +123,6 @@ fn build_ignore_set(
                 "build",
                 "dist",
                 "node_modules",
-                ".wasmbuild",
                 ".next",
                 ".nuxt",
                 "bin",
@@ -416,17 +379,13 @@ impl CliCommand for DevCommand {
         let (reload_tx, mut reload_rx) = mpsc::channel::<()>(1);
 
         // Build initial ignore set including artifact path and project-specific build directories
-        let initial_ignore_set = build_ignore_set(
-            &canonical_project_root,
-            Some(&artifact_path),
-            &project_context,
-        );
+        let initial_ignore_set = build_ignore_set(&canonical_project_root, &project_context);
         debug!(
             ignore_count = initial_ignore_set.len(),
             project_type = ?project_context,
             "built initial file watcher ignore set"
         );
-        let ignore_paths = Arc::new(RwLock::new(initial_ignore_set));
+        let ignore_paths = Arc::new(initial_ignore_set);
         let ignore_paths_notify = ignore_paths.clone();
 
         let canonical_project_root_notify = canonical_project_root.clone();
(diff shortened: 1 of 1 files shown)


navahas:
I used the tests as handy smoke checks while developing. I kept them in the PR in case others want to run them locally, feel free to remove or modify them.

brooksmtownsend:
Just a note @navahas im fixing the windows test issue in #36, so don't worry about that test specifically

crates/wash/src/cli/dev.rs
```diff
@@ -47,6 +47,140 @@ use crate::{
...
+    // Common directories for all project types
+    let mut dirs_to_ignore = vec![".git", ".wash"];
+
+    // Add project-type-specific ignore patterns
+    match project_context {
+        ProjectContext::Rust { .. } => {
+            dirs_to_ignore.extend_from_slice(&["target", ".wasmbuild"]);
+        }
+        ProjectContext::TypeScript { .. } => {
+            dirs_to_ignore.extend_from_slice(&["node_modules", "dist", "build", ".next", ".nuxt"]);
```
brooksmtownsend, line 93: Is `.nuxt` a typescript framework? For `.next`, does it ever include metadata that we should rebuild on?

crates/wash/src/cli/dev.rs
```diff
@@ -47,6 +47,140 @@ use crate::{
...
+        }
+    }
+
+    // Common directories for all project types
+    let mut dirs_to_ignore = vec![".git", ".wash"];
+
+    // Add project-type-specific ignore patterns
+    match project_context {
+        ProjectContext::Rust { .. } => {
+            dirs_to_ignore.extend_from_slice(&["target", ".wasmbuild"]);
```
brooksmtownsend: I think we can remove .wasmbuild here

crates/wash/src/cli/dev.rs
```diff
@@ -47,6 +47,140 @@ use crate::{
...
+                    ".nuxt",
+                ]);
+            }
+            if detected_types.iter().any(|t| t == "Go") {
+                dirs_to_ignore.extend_from_slice(&["bin", "pkg", "vendor"]);
+            }
+        }
+        ProjectContext::General => {
+            // For general context, include common patterns from all project types
+            dirs_to_ignore.extend_from_slice(&[
```
brooksmtownsend, line 121: same comment re: removing .wasmbuild/.nuxt

crates/wash/src/cli/dev.rs
```diff
@@ -47,6 +47,140 @@ use crate::{
...
+    let mut ignore_paths = HashSet::new();
+
+    // Add the artifact path's parent directory if provided
+    if let Some(artifact) = artifact_path {
+        if let Some(artifact_dir) = artifact.parent() {
+            if let Ok(canonical) = artifact_dir.canonicalize() {
+                ignore_paths.insert(canonical);
+            }
+        }
+    }
```
brooksmtownsend: Hrm, what happens if we're outputting the artifact directly into the source of the project?

brooksmtownsend: I think that this is a good heuristic, but I think it may be best to leave it out and err on the side of an accidental rebuild once in a while

crates/wash/src/cli/dev.rs
```diff
@@ -47,6 +47,140 @@ use crate::{
...
+
+    // Common directories for all project types
+    let mut dirs_to_ignore = vec![".git", ".wash"];
```
brooksmtownsend: I actually think if the config changes in `.wash`, we _do_ want to rebuild since it may affect the build

crates/wash/src/cli/dev.rs
```diff
@@ -540,3 +724,224 @@ pub async fn handle_request<'a>(
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
brooksmtownsend, line 695: tests 😍

crates/wash/src/cli/dev.rs
```diff
@@ -250,14 +384,41 @@ impl CliCommand for DevCommand {
...
+        );
+        debug!(
+            ignore_count = initial_ignore_set.len(),
+            project_type = ?project_context,
+            "built initial file watcher ignore set"
+        );
+        let ignore_paths = Arc::new(RwLock::new(initial_ignore_set));
+        let ignore_paths_notify = ignore_paths.clone();
+
+        let canonical_project_root_notify = canonical_project_root.clone();
```
brooksmtownsend: I might be missing some rationale here, but I think that removing the RwLock and not worrying about inserting an artifact path later on would make this quite a bit simpler since I think the artifact path should be nested under a path we ignore already

crates/wash/src/cli/dev.rs
```diff
@@ -47,6 +47,140 @@ use crate::{
...
+        if canonical_path.starts_with(ignore_path) {
+            return true;
+        }
+    }
+
+    // Also check relative paths from canonical project root
+    if let Ok(relative_path) = canonical_path.strip_prefix(canonical_project_root) {
+        // Check if any component of the relative path matches common build directories
+        for component in relative_path.components() {
+            if let Some(component_str) = component.as_os_str().to_str() {
```
brooksmtownsend: I think I'd like to skip this loop / recursive checking for the path components since we are providing a set of ignored paths already to this function, and that would help eliminate some confusing edge cases

brooksmtownsend (changes requested):
I had a lot of comments for this one @navahas but this looks really nice! My requests are all around removing some of the logic that might ignore more than we want, I'd rather trigger a couple extra rebuilds than ignore a legitimate change.

If you ran into anything when developing this, please feel free to push back on my suggestions

crates/wash/src/cli/dev.rs
navahas: .nuxt is the build output from Nuxt (Vue-based, JS/TS). Similar to .next for Next.js, it only has generated code/metadata, so safe to ignore.

crates/wash/src/cli/dev.rs
navahas: Ah, right! Forgot to remove on my end. It was an initial test placeholder for possible new build directories from future frameworks I might have missed in this PR. Safe to drop it.

crates/wash/src/cli/dev.rs
navahas: Agree, an occasional extra rebuild is safer than missing one.

crates/wash/src/cli/dev.rs
brooksmtownsend: Sounds good 😃

crates/wash/src/cli/dev.rs
navahas: Fair, recursion is something I struggle to agree with too, but here I wasn’t sure how else to reliably catch nested cases like subproject/target without listing every possible nested path in the ignore set. That was the main reason I added it, though I get it adds edge cases.

navahas:
> I had a lot of comments for this one @navahas but this looks really nice! My requests are all around removing some of the logic that might ignore more than we want, I'd rather trigger a couple extra rebuilds than ignore a legitimate change.
> 
> If you ran into anything when developing this, please feel free to push back on my suggestions

Appreciate! Honestly, I’m just happy to contribute with an initial PR here, I’m still getting familiar with the whole project, and you’ve got a much better sense of the bigger picture. I’m happy for any ideas or modifications you think fit better.

navahas:
## Manual Verification

### Terminal A
In the component folder
```wash dev .```

### Terminal B

#### Real source change
- Terminal A (wash dev): SHOULD rebuild
```bash
echo "// change $(date)" >> src/lib.rs
```

#### Ignored dirs
- Terminal A (wash dev): SHOULD NOT rebuild
```bash
mkdir -p target/tmp node_modules/tmp
touch target/tmp/as.txt node_modules/tmp/cs.txt
```

#### Relevant dirs
- Terminal A (wash dev): SHOULD rebuild
```bash
mkdir -p dist/tmp_2
touch dist/tmp_2/bs.txt
```
```bash
mkdir -p .wash/tmp_dir
touch .wash/tmp_dir/log.txt
```

#### Similar name, not ignored
- Terminal A: SHOULD rebuild
```bash
mkdir -p targetx
touch targetx/should-trigger.txt
```

brooksmtownsend (approved):
LGTM @navahas ! Looks like you just need a rebase

navahas:
sorted!

PR #44, opened 2025-08-22
fix(plugin): improve error handling for command argument parsing in TestCommand

### LINKED ISSUE #12: Fix clap help text exiting integration tests
## Description

The plugin integration test has commented-out code for testing the `--help` flag because clap's help text causes the test to exit unexpectedly. This needs to be resolved so the help functionality can be properly tested.

## Location

`tests/integration_plugins.rs:43` in the `test_plugin_test_oauth_comprehensive` test function

## Current Behavior

The test code that would verify `--help` functionality is commented out because:
- Clap's help text calls `std::process::exit()`
- This causes the entire test to exit instead of returning normally
- The test cannot validate help text output as intended

## Proposed Solution

Options to resolve this:
1. **Separate test process**: Run the help test in a separate process/subprocess
2. **Mock/hook mechanism**: Hook into clap's exit behavior during tests
3. **Conditional exit**: Make the exit behavior conditional in test mode
4. **Test isolation**: Move help tests to a dedicated integration test file

## Context

This affects the ability to test plugin help functionality comprehensively. The commented code shows the intended test structure:

```rust
let test_cmd_basic = TestCommand {
    plugin: oauth_plugin_path.clone(),
    args: vec!["--help".to_string()],
    hooks: vec![],
};
```

jtakakura:
## Feature or Problem

This PR is to fix the clap help text issue in the plugin integration test.

Previously, the test for the `--help` flag was commented out because it would cause the test to exit unexpectedly. This PR resolves the issue by handling the `--help` flag correctly, allowing the test to run and pass as expected.

This change ensures that the `--help` functionality is properly tested and validated.

## Related Issues

Fixes #12

### DIFF
--- a/tests/integration_plugins.rs
+++ b/tests/integration_plugins.rs
@@ -40,32 +40,35 @@ async fn test_plugin_test_inspect_comprehensive() -> Result<()> {
         inspect_plugin_path.display()
     );
 
-    // TODO(#12): Weirdly, clap's help text will actually _exit_ the test if this
-    // runs. So it either needs to be tested separately or hooked into somehow
     // Test 1: Basic plugin test without any command or hook flags
-    // eprintln!("🔍 Test 1: Basic plugin test (help)");
-    // let test_cmd_basic = TestCommand {
-    //     plugin: oauth_plugin_path.clone(),
-    //     args: vec!["--help".to_string()],
-    //     hooks: vec![],
-    // };
-    // let plugin_cmd_basic = PluginCommand::Test(test_cmd_basic);
+    eprintln!("🔍 Test 1: Basic plugin test (help)");
 
-    // let result_basic = plugin_cmd_basic
-    //     .handle(&ctx)
-    //     .await
-    //     .context("Failed to execute basic plugin test")?;
+    let oauth_plugin_path = PathBuf::from("plugins/oauth");
+
+    let test_cmd_basic = TestCommand {
+        plugin: oauth_plugin_path.clone(),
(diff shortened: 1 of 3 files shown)


tests/integration_plugins.rs
```diff
@@ -40,32 +40,35 @@ async fn test_plugin_test_inspect_comprehensive() -> Result<()> {
...
-    //     !result_basic.is_success(),
-    //     "Basic plugin test should succeed"
-    // );
-    // // The description
-    // assert!(result_basic.text().contains(
-    //     "OAuth2 server for authentication"
-    // ));
-    // eprintln!("✅ Basic plugin test passed");
+    assert!(
+        !result_basic.is_success(),
```
jtakakura, line 62: @brooksmtownsend  Currently, the test expects `--help` to fail (`is_success()` is false). I can either:

1.  Keep the current behavior where it exits with an error code, so the test still passes.
2.  Change the code to handle the `--help` flag gracefully (making `is_success()` true), and update the test to reflect this.

Which approach do you think is better for the project?

brooksmtownsend: I think we should keep the behavior in the code as-is, don't want to make modifications there for the test. I think that's okay for it to expect the code to fail, as long as we leave a comment as to why 😄

jtakakura: Thanks for the feedback! I'll keep the current behavior.
I've changed the PR to be ready for review now.

brooksmtownsend (approved):

PR #68, opened 2025-09-03
fix/template locality

brooksmtownsend:
- **fix(config): store templates in global not project**
- **release(wash): v1.0.0-beta.5**

## Feature or Problem
This PR fixes an issue with deserializing templates that ensures that missing templates don't throw an error and instead default to an empty list. It also adds the `--global` flag to the `wash config init` command which ensures that we're initializing configuration in the local project context by default.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5466,7 +5466,7 @@ dependencies = [
 
 [[package]]
 name = "wash"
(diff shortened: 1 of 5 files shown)


brooksmtownsend:
supersedes  #67

ricochet (approved):

Issue #23, opened 2025-08-13
Include color output in build error messages during dev mode


brooksmtownsend:
## Description

When component builds fail during development mode, the error output doesn't include color formatting. This makes it harder to read and parse error messages compared to running builds directly.

## Location

`crates/wash/src/cli/dev.rs:345` in the rebuild error handling

## Current Behavior

```rust
Err(e) => {
    info!("failed to build component, will retry on next file change");
    // TODO(GFI): This doesn't include color output
    // This nicely formats the error message
    error!("{e}");
    // ...
}
```

The error output lacks color formatting that would normally be present when running build tools directly.

## Proposed Solution

Preserve color formatting in error output:

1. **Capture colored output**: Ensure build processes preserve terminal color codes
2. **Pass through colors**: Don't strip ANSI color codes when displaying errors
3. **Terminal detection**: Respect user's color preferences and terminal capabilities

## Example Implementation

```rust
Err(e) => {
    info!("failed to build component, will retry on next file change");
    
    // Preserve color output if terminal supports it
    if atty::is(atty::Stream::Stderr) {
        // Display with colors preserved
        eprintln!("{}", e);
    } else {
        // Fallback to plain text for non-terminals
        error!("{e}");
    }
    // ...
}
```

## Benefits

- **Better readability**: Color-coded errors are easier to scan and understand
- **Consistent experience**: Matches what users see when running builds directly
- **Error highlighting**: Syntax errors, warnings, and suggestions are properly highlighted

## Implementation Notes

- Should respect `NO_COLOR` environment variable
- Consider using `termcolor` or similar crate for cross-platform support
- May need to configure build tools to output colors even when not in TTY mode

brooksmtownsend:
Would recommend holding off until https://github.com/wasmCloud/wash/pull/73 merges to take this one on since there's quite a few changes in that branch!

lferrazzini95:
Hi @brooksmtownsend this can be closed right?

PR #70, opened 2025-09-08
feat: add color output in build error messages

lferrazzini95:
## Feature or Problem
As described in [issue 23](https://github.com/wasmCloud/wash/issues/23) build errors occuring with the dev command do not output any colors. The approach taken here is to first check if atty and then depending on `NO_COLOR` is set the environment variable `CARGO_TERM_COLOR` is set to `always`.

## Related Issues
[Issue 23](https://github.com/wasmCloud/wash/issues/23)

## Release Information
`next`

## Consumer Impact
If no color output for dev build messages is desired `NO_COLOR=1` needs to be set.

## Testing
No tests were added. The output was tested manually.

### Unit Test(s)
None

### Acceptance or Integration
None

### Manual Verification
Ran wash dev /path/to/example/component with `NO_COLOR=1`, `NO_COLOR=0` and NO_COLOR not set. The component contained build errors and the colors were visible only when `NO_COLOR=0` or `NO_COLOR` was not set.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -353,6 +353,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi 0.1.19",
+ "libc",
+ "winapi",
+]
+
+[[package]]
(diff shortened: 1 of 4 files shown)


brooksmtownsend (changes requested):
Hey @lferrazzini95 , thank you for the PR! I was wondering if you were able to find any alternatives to introducing a block of `unsafe` code to set the environment variable, especially because people may have things like `CARGO_TERM_COLOR=none` set in their env and we should respect that.

lferrazzini95:
Hi @brooksmtownsend, thanks a lot for your review and the remarks! First PR for me so very thankful for any inputs!
I took another approach to set the `--color` flag as a cargo build argument. This way if I am not missing anything the cases you mentioned should be covered and also the unsafe block is gone. I hope this is a better approach :)
Best
Luca

brooksmtownsend (approved):

PR #79, opened 2025-09-24
refactor: rename artifact-path to component-path

ricochet:
Renames `artifact_path` to a more descriptive name, `component_path`.

We always expect a `.wasm`, the name `artifact_path` feels like it implies
it could be a different output format like a OCI tarball.

### DIFF
--- a/.wash/config.json
+++ b/.wash/config.json
@@ -1,6 +1,6 @@
 {
   "build": {
-    "artifact_path": "target/debug/wash",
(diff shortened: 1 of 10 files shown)


ricochet:
Might be a spicy 🌶️ opinion so would like for @brooksmtownsend to be back from vacation 🌴 to weight in.

brooksmtownsend (approved):

PR #81, opened 2025-09-25
fix: make search slightly for flexible for custom

ricochet:


MendyBerger (approved):
You have a lint error. But rest looks good to me

PR #85, opened 2025-09-30
feat: add non-interactive mode

ricochet:

### DIFF
--- a/crates/wash/src/cli/mod.rs
+++ b/crates/wash/src/cli/mod.rs
@@ -74,6 +74,7 @@ pub trait CliCommandExt: CliCommand {
                     trace!(?hook, ?hook_type, "executing pre-hook for command");
                     let mut data = Ctx::builder(hook.metadata.name.clone())
                         .with_background_processes(ctx.background_processes.clone())
+                        .skip_confirmation(ctx.is_non_interactive())
                         .build();
                     // TODO(IMPORTANT): context about the command and runner
                     let runner = data
@@ -115,6 +116,7 @@ pub trait CliCommandExt: CliCommand {
                     trace!(?hook, "executing post-hook for command");
                     let mut data = Ctx::builder(hook.metadata.name.clone())
                         .with_background_processes(ctx.background_processes.clone())
+                        .skip_confirmation(ctx.is_non_interactive())
                         .build();
                     // TODO(IMPORTANT): context about the command and runner
                     let runner = data
@@ -276,6 +278,8 @@ pub struct CliContext {
     /// Stores the handles to background processes spawned by host_exec_background. We want to
     /// constrain the processes spawned by components to the lifetime of the CLI context.
     background_processes: Arc<RwLock<Vec<Child>>>,
+    /// Whether to run in non-interactive mode (skip terminal checks for host exec)
+    non_interactive: bool,
 }
 
 #[cfg(unix)]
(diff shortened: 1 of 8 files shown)


crates/wash/src/cli/mod.rs
```diff
@@ -297,7 +301,7 @@ impl Deref for CliContext {
 
 impl CliContext {
     /// Creates a new [CliContext] with the specified output kind and directory paths.
-    pub async fn new() -> anyhow::Result<Self> {
+    pub async fn new(non_interactive: bool) -> anyhow::Result<Self> {
```
vados-cosmonic: We should probably go with a builder here if this needs to be modifiable, or use the default? We could have a builder if we want it to be editable.

crates/wash/src/runtime/mod.rs
```diff
@@ -68,6 +71,7 @@ impl CtxBuilder {
                 .build(),
             runtime_config: None,
             background_processes: None,
+            non_interactive: false,
```
vados-cosmonic: I'm surprised this has to get passed into the runtime level? The runtime shouldn't need to know that the CLI was set up to be non interactive...

Can we store this as runtime_config instead?

ricochet: I expect runtime-config to be more for the component data rather than host data. I agree it's pretty bizarro. Makes me think the host_exec approach needs to be rethought into a capability system that's loaded here and then we have skip_checks as a setting. For now I'm going to rename this on the runtime ctx to skip_checks so that cli info isn't leaking.

crates/wash/src/cli/mod.rs
```diff
@@ -295,9 +299,34 @@ impl Deref for CliContext {
...
 }
 
+/// Builder for constructing a CliContext
+pub struct CliContextBuilder {
+    non_interactive: bool,
+}
+
+impl CliContextBuilder {
+    /// Set whether to run in non-interactive mode
+    pub fn with_non_interactive(mut self, non_interactive: bool) -> Self {
```
vados-cosmonic:
```suggestion
    pub fn non_interactive(mut self, non_interactive: bool) -> Self {
```

crates/wash/src/cli/mod.rs
```diff
@@ -295,9 +299,34 @@ impl Deref for CliContext {
...
+        self.non_interactive = non_interactive;
+        self
+    }
+
+    /// Build the CliContext
+    pub async fn build(self) -> anyhow::Result<CliContext> {
+        CliContext::create(self.non_interactive).await
```
vados-cosmonic: This should be a `CliContext { ... }` creation probably... or maybe you can do the original `new()` and then modify it afterwards? Either way this is usually not what the `build()` would look like...

crates/wash/src/cli/mod.rs
```diff
@@ -295,9 +299,34 @@ impl Deref for CliContext {
...
+}
+
 impl CliContext {
-    /// Creates a new [CliContext] with the specified output kind and directory paths.
-    pub async fn new() -> anyhow::Result<Self> {
+    /// Creates a new CliContext builder with default settings
+    pub fn builder() -> CliContextBuilder {
+        CliContextBuilder {
+            non_interactive: false,
+        }
```
vados-cosmonic:
```suggestion
        CliContextBuilder::default()
```

crates/wash/src/cli/mod.rs
```diff
@@ -295,9 +299,34 @@ impl Deref for CliContext {
...
+    }
+
+    /// Internal method to create a CliContext with the given settings
+    async fn create(non_interactive: bool) -> anyhow::Result<Self> {
```
vados-cosmonic: can we go back to `new()` having no args?

brooksmtownsend (approved):
This looks nice to me and leaves good room for us to add in other ways to skip confirmation, like trusting certain plugins!

Issue #78, opened 2025-09-24
[FEATURE] Use wasm-metadata to populate annotations for oci push


ricochet:
This is part of supporting annotations for the push command, but is a specific and necessary initial step, so breaking out as a sub-issue of wasmCloud/wasmCloud#4932. 

When running `wash oci push` with a component, the annotations added in custom section of the wasm should be preserved and mapped to OCI annotations.

The BA tool `wkg` already does this and so should we.  From a design perspective, we are aiming to consolidate this type of logic into wasm-tools so that many embedding implementations share this capability. That should mostly be the case today by using the wasm-metadata crate to read these: https://github.com/bytecodealliance/wasm-tools/tree/main/crates/wasm-metadata/src/oci_annotations

See how cargo component implements this: https://github.com/bytecodealliance/cargo-component/pull/372/files

f4z3r:
I might give this a go today

ricochet:
Fixed by wasmCloud/wash#86

PR #53, opened 2025-08-27
feat(oci): add annotations as optional parameters to OCI push

brooksmtownsend:
## Summary
- Adds `--annotation key=value` parameter (repeatable) for custom OCI annotations
- Includes convenience parameters for common OpenContainer standards (`--description`, `--source`, `--url`, `--version`, `--licenses`, `--author`)
- Updates push command output to display component digest

## Test plan
- [x] All existing tests pass (41/41)
- [x] 17 new unit tests added covering annotation parsing and manifest creation
- [x] Code formatted with `cargo +nightly fmt`
- [x] No clippy warnings
- [x] Manual testing with various annotation combinations

Closes wasmCloud/wasmCloud#4932

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #86, opened 2025-10-06
Add OCI annotations based on metadata attached to wasm payload

f4z3r:
When pushing an OCI artifact, attach metadata such as source, url, description, version, and licenses to the container image based on the metadata attached to the wasm payload that was constructed in the build step.

## Feature or Problem

See #78 

## Related Issues

#78

## Release Information

Next.

## Consumer Impact

OCI artifacts now contain metadata when pushed to a registry.

### Unit Test(s)

TODO

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,12 +62,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e9d4ee0d472d1cd2e28c97dfa124b3d8d992e10eb0a035f33f5d12e3a177ba3b"
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -364,6 +358,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "auditable-serde"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5c7bf8143dfc3c0258df908843e169b5cc5fcf76c7718bd66135ef4a9cd558c5"
+dependencies = [
+ "semver",
+ "serde",
+ "serde_json",
+ "topological-sort",
+]
+
+[[package]]
 name = "autocfg"
 version = "1.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


f4z3r:
@ricochet was that what you were thinking? I wanted to quickly get your feedback before I start implementing some tests, as I am not 100% sure how to tackle unit testing for scenarios where I need a build artifact. I would also add additional annotations based on the remaining meatadata available on wasm payloads.

PS: the code changes in the OCI push logic I 1-to-1 copied from PR #53 (thanks @brooksmtownsend 🤣 ).

ricochet:
Thank you for taking this! Your approach looks right to me. I'm little surprised the underlying crates are missing helpers like the oci client not having a creator with annotations and wasm-tools today doesn't have the canonical oci labels like a metadata2oci helper, but this is something we could upstream from wash later. I'm not asking for changes, what you have here looks correct to me.

Little heads up that `author` will change to authors in an upcoming wasm-tools release:
https://github.com/bytecodealliance/wasm-tools/pull/2334/files

f4z3r:
Sweet, thanks for the update. I will have a look to see if there is really no utility function to generate the labels.

Thanks for the heads-up regarding the renaming of `author`. Then I will have a look later at implementing some tests, adding the remaining labels, and potentially having a look at what upstreaming the changes would entail 👍🏽

f4z3r:
Alright, I added the remaining annotations that are in the intersection of [`Metadata` fields](https://docs.rs/wasm-metadata/latest/wasm_metadata/struct.Metadata.html) and [pre-defined keys in the opencontainer spec](https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys). I also added an annotation for the creation time of the OCI artifact.

I have two questions:

1. I am very unsure how to test this, as I would essentially need a build artifact to test this. Assuming that `wasm-tools` works (IMO we should not test that), there is little to no logic that is actually implemented in the annotation mapping. Do you see any sensible way of adding tests for this, or would you say it is not necessary?
2. The logic in `crates/wash/src/oci.rs` is a different story. There I copied Brooks' tests over. However, some tests like `test_standard_opencontainer_annotations` seem to test whether HashMaps work? Or am I missing something obvious here? I would have removed these personally.

ricochet:
> Alright, I added the remaining annotations that are in the intersection of [`Metadata` fields](https://docs.rs/wasm-metadata/latest/wasm_metadata/struct.Metadata.html) and [pre-defined keys in the opencontainer spec](https://specs.opencontainers.org/image-spec/annotations/#pre-defined-annotation-keys). I also added an annotation for the creation time of the OCI artifact.
> 
> I have two questions:
> 
> 1. I am very unsure how to test this, as I would essentially need a build artifact to test this. Assuming that `wasm-tools` works (IMO we should not test that), there is little to no logic that is actually implemented in the annotation mapping. Do you see any sensible way of adding tests for this, or would you say it is not necessary?
> 2. The logic in `crates/wash/src/oci.rs` is a different story. There I copied Brooks' tests over. However, some tests like `test_standard_opencontainer_annotations` seem to test whether HashMaps work? Or am I missing something obvious here? I would have removed these personally.

Your tests seem largely sufficient to me although I am very tempted to follow the Beyoncé rule for testing wasm-tools Metadata field stability, although easy to argue we're mostly safe via types. Upstreaming which OCI annotations relate to which metadata fields is ultimately the best longterm defense.

brooksmtownsend (approved):
Looks really nice!! Very simple.

PR #90, opened 2025-10-06
ci: update installer workflow to test on change

ricochet:


brooksmtownsend (approved):
I noticed that we get rate limited pretty quickly here, is that going to cause problems in tests?

ricochet:
I'm thinking it'll be pretty rare for these to run since I'm not aware of anymore enhancements we want to make to the installer

Issue #25, opened 2025-08-13
Refactor CliContext to use Arc-ed trait object for app strategy


brooksmtownsend:
## Description

The `CliContext` struct currently uses conditional compilation to store either `Xdg` or `Windows` strategy types directly. This can be simplified by using an Arc-ed trait object, which would eliminate the need for conditional compilation and reduce code duplication.

## Location

`crates/wash/src/cli/mod.rs:259` in the `CliContext` struct definition

## Current Behavior

```rust
#[derive(Debug, Clone)]
pub struct CliContext {
    // TODO(GFI): Just store an Arc-ed trait object
    #[cfg(unix)]
    app_strategy: Xdg,
    #[cfg(windows)]
    app_strategy: Windows,
    // ... other fields
}
```

This requires separate `Deref` implementations for each platform and conditional compilation throughout the codebase.

## Proposed Solution

Replace the platform-specific fields with a trait object:

```rust
// Define a common trait for directory strategies
trait DirectoryStrategy: Debug + Send + Sync {
    fn config_dir(&self) -> PathBuf;
    fn data_dir(&self) -> PathBuf;
    fn cache_dir(&self) -> PathBuf;
    // ... other common methods
}

// Implement the trait for existing types
impl DirectoryStrategy for Xdg { /* ... */ }
impl DirectoryStrategy for Windows { /* ... */ }

#[derive(Debug, Clone)]
pub struct CliContext {
    app_strategy: Arc<dyn DirectoryStrategy>,
    // ... other fields
}
```

## Benefits

- **Eliminates conditional compilation**: Single implementation regardless of platform
- **Cleaner code**: No need for duplicate `Deref` implementations
- **Better testability**: Easier to mock directory strategies for testing
- **Extensibility**: Could support custom directory strategies in the future

## Implementation Notes

- The `directories` crate types (`Xdg`, `Windows`) would need wrapper types to implement the trait
- Consider using `Box<dyn DirectoryStrategy>` if `Arc` isn't needed for this specific case
- Ensure `Debug` and `Clone` implementations work correctly with the trait object

PR #96, opened 2025-10-08
Use trait in CliContext to avoid conditional compilation

f4z3r:
## Feature or Problem

Avoid conditionally compiling based on target OS to get AppStrategy. Instead use wrapper trait to provide same functionality in a way that makes it simpler to test and extend strategy capabilities.

## Related Issues

#25

## Release Information

Next.

## Consumer Impact

None.

### DIFF
--- a/crates/wash/src/cli/mod.rs
+++ b/crates/wash/src/cli/mod.rs
@@ -1,16 +1,15 @@
 //! The main module for the wash CLI, providing command line interface functionality
 
+use std::path::PathBuf;
 use std::{ops::Deref, path::Path, sync::Arc};
 
 use anyhow::Context as _;
-use etcetera::{AppStrategy as _, AppStrategyArgs, choose_app_strategy};
 use tokio::{process::Child, sync::RwLock};
 use tracing::info;
 
-#[cfg(windows)]
 use etcetera::app_strategy::Windows;
-#[cfg(unix)]
 use etcetera::app_strategy::Xdg;
+use etcetera::{AppStrategy, AppStrategyArgs, choose_app_strategy};
 
 use serde_json::json;
 use tracing::{debug, error, instrument, trace};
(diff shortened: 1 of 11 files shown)


crates/wash/src/cli/mod.rs
```diff
@@ -259,17 +258,91 @@ impl CommandOutput {
     }
 }
 
+/// Non-generic copy of etcetera::AppStrategy trait to make it dyn-compatible
+pub trait DirectoryStrategy: std::fmt::Debug + Send + Sync {
```
f4z3r, line 262: I modeled the methods to match one to one the ones from `AppStrategy`, with the exception to avoid generic methods to make it dyn-compatible. This avoids having to change the code at the call-sites.

crates/wash/src/cli/mod.rs
```diff
@@ -259,17 +258,91 @@ impl CommandOutput {
...
 /// or a custom configuration if needed.
 #[derive(Debug, Clone)]
 pub struct CliContext {
-    // TODO(#25): Just store an Arc-ed trait object
-    #[cfg(unix)]
-    app_strategy: Xdg,
-    #[cfg(windows)]
-    app_strategy: Windows,
+    /// Application strategy to access configuration directories.
+    app_strategy: Arc<dyn DirectoryStrategy>,
```
f4z3r, line 345: This indeed requires an `Arc` as `Xdg` and `Windows` do not implement `Clone`. So using a `Box` would require some wrapper object/enum with smart `Clone`ing implemented.

crates/wash/src/cli/mod.rs
brooksmtownsend: Well shoot, the provided methods being non-dyn compatible is a bit annoying! But at least this solves it fairly well.

crates/wash/src/cli/mod.rs
brooksmtownsend: Nice!

brooksmtownsend (approved):
Very nice! Will be great to remove the conditional compilation here. Disappointing that the `AppStrategy` trait isn't dyn compatible

Issue #16, opened 2025-08-13
Add comprehensive testing for CLI args config override behavior


brooksmtownsend:
## Description

The configuration loading system needs more comprehensive testing to ensure that CLI arguments properly override existing configuration values without replacing present values with empty values.

## Location

`crates/wash/src/config.rs:153` in the `load_config` function

## Current Behavior

The `load_config` function implements a hierarchical configuration system with this precedence:
1. Default values
2. Global config (`~/.wash/config.json`)
3. Local project config (`.wash/config.json`)
4. Environment variables (`WASH_` prefix)
5. Command line arguments (highest priority)

However, there's incomplete testing to ensure CLI args behave correctly when overriding existing configuration.

## Problem

The current implementation may have issues where:
- CLI args with empty/default values might replace valid configured values
- Partial CLI config objects might override entire config sections
- Edge cases in the merge behavior aren't properly tested

## Proposed Solution

Add comprehensive test coverage for:

1. **Selective override testing**: Verify CLI args only override specified fields
2. **Empty value handling**: Ensure empty CLI values don't replace valid config values  
3. **Nested config merging**: Test deep merging behavior for complex config structures
4. **Edge cases**: Test various combinations of config sources

## Example Test Cases

```rust
#[test]
fn test_cli_args_selective_override() {
    // Verify that setting only one CLI arg doesn't clear other configured values
}

#[test] 
fn test_cli_args_empty_values() {
    // Verify empty CLI values don't override existing valid config
}
```

## Context

This is critical for ensuring users can rely on CLI arguments working as expected without inadvertently breaking their saved configuration.

PR #97, opened 2025-10-08
Add testing suite for configuration precedence and omission testing

f4z3r:
## Related Issues

#16

## Consumer Impact

None

### DIFF
--- a/crates/wash/src/config.rs
+++ b/crates/wash/src/config.rs
@@ -148,7 +148,7 @@ where
     }
 
     // Environment variables with WASH_ prefix
-    figment = figment.merge(Env::prefixed("WASH_").split("_"));
+    figment = figment.merge(Env::prefixed("WASH_").split("__"));
 
     // TODO(#16): There's more testing to be done here to ensure that CLI args can override existing
     // config without replacing present values with empty values.
@@ -320,10 +320,10 @@ mod test {
         local_config.build = Some(BuildConfig {
             rust: Some(RustBuildConfig {
                 release: true,
-                ..RustBuildConfig::default()
+                ..Default::default()
             }),
-            ..BuildConfig::default()
-        }); // should override global
+            ..Default::default()
+        }); // Should override global
         local_config.templates = Vec::new(); // should take templates from global
         save_config(&local_config, &local_config_file).await?;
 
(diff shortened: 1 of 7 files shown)


crates/wash/src/config.rs
```diff
@@ -148,7 +148,7 @@ where
     }
 
     // Environment variables with WASH_ prefix
-    figment = figment.merge(Env::prefixed("WASH_"));
+    figment = figment.merge(Env::prefixed("WASH_").split("_"));
```
f4z3r: I am not quite done with the testing, but wanted to make sure this is desired. I have the feeling this is the typical way to do this, instead of using the default from `figment` (which is a dot - `.`).

Not setting this does not allow configurations via the env vars as shown in the test below (https://github.com/wasmCloud/wash/pull/97/files#diff-f7f9d11319f33b9af91da4ef6ce4b7111c1baceef1b82db506c4612ba6f808b1R356). If this is removed, the test fails as the splitting does not work on underscores.

f4z3r: I will need to add a test on how this works on "multi-word" configuration options though such as here: https://github.com/wasmCloud/wash/blob/0d7cde20b51e8f253b655864c8f68cdb7cfeaee9/crates/wash/src/component_build.rs#L48

f4z3r: I updated this to use two underscored (`__`) as the separator. Otherwise multi-word configurations cannot be parsed from the environment. This results in the following sample configuration via environment variables:

```sh
WASH_BUILD__RUST__CUSTOM_COMMAND="[cargo,build]"
```

The only slightly weird stuff is the single underscore in the prefix. We could update that to a double underscore as well to have it be consistent. Or do you see any other separator instead of double underscore that might make sense (note single underscore will not work properly) @brooksmtownsend ? The figment default is a dot, but this is not supported properly in many shells AFAIK.

f4z3r:
Is it possible that the tests are slightly flaky? I am unsure I understand why the tests on Ubuntu fail. Regarding the tests on Windows, this is extremely weird. While the test CLI context creates a new temporary directory for every test that should host all files, it seems that on Windows some configurations are found in the test CLI context...?

automation-wasmcloud:

PR #120, opened 2025-11-06
feat!: use wasi config rc.1

ricochet:


examples/http-counter/src/lib.rs
```diff
@@ -1,52 +1,71 @@
 use anyhow::{Context, Result};
-use wasmcloud_component::http::{self, StatusCode};
```
ricochet, line 2: Had an annoying time with wit conflicts in wasm_component crate that I yanked it and am just using direct bindings.

lxfontes (approved):

PR #133, opened 2025-11-07
refactor: handle bind failure

### LINKED ISSUE #128: `HostApi::on_workload_bind` can leak plugin internals if another binding fails
Minor footgun but the impl right now bails when a binding fails without unbinding other bound plugins. Implementations could use a  plugin internal queue across `on_workload_bind` and `on_workload_resolved` to handle this but that seems like accidental boilerplate to me.

ricochet:
Refactors on_workload_unbind to use workload id as all
cases and callers only every used id of the workload and
allow unbind to use UnResolvedWorkloads as well as Resolved.

The runtime ensures proper cleanup automatically in LIFO order

Fixes #128

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -25,6 +25,14 @@ use crate::{
     wit::{WitInterface, WitWorld},
 };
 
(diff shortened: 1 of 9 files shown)


lxfontes (approved):
👍 
this helps enforce that plugins should store contextual data by workload and not component id.

PR #150, opened 2025-11-09
feat(dev): add in-memory plugin

ricochet:


lxfontes (approved):

lxfontes:
Tested with examples/http-counter

PR #156, opened 2025-11-14
feat: Bundle HTTP Implementation with Runtime

lxfontes:
HTTP is tightly coupled to the runtime via wasmtime-wasi-http, complicating integration via plugins. The deciding point seems to be: Does the WASI interface require a special trait in the Component Context? If the answer is Yes, then it shouldn't be a plugin.

The approach in this PR allows for 2 integration points:
- `http::HostHandler`: wash-runtime consumers looking to implement their own HTTP Ingress/Egress (transport)
- `http::Router`: wash-runtime consumers implementing custom HTTP Logic or filtering logic (routing), while using the provided transport

~As this is a fairly substantial change, I will not remove the wasi_http implementations from `plugin` and `washlet` just yet.~

Removed wasi_http from both plugin and washlet as they would simply not work correctly anymore.


crates/wash-runtime/src/host/http.rs
lxfontes, line 1: this is where the duplicated code from plugins/washlet wasi_http.rs went

crates/wash-runtime/src/host/http.rs
```diff
@@ -250,37 +411,41 @@ impl HostPlugin for HttpServer {
...
+    async fn on_workload_unbind(&self, workload_id: &str) -> anyhow::Result<()> {
         // Remove from workload handles
         let mut handles_guard = self.workload_handles.write().await;
-        handles_guard.retain(|_, (handle, _, _)| handle.id() != workload_id);
+        handles_guard.remove(workload_id);
+        drop(handles_guard);
 
-        // Remove from workload configs
-        self.workload_configs.write().await.remove(workload_id);
+        self.router.on_workload_unbind(workload_id).await?;
```
ricochet: Workload handles are removed before notifying the router, creating a window where requests could be routed to a missing workload.

```suggestion
        // Remove from workload handles
        self.router.on_workload_unbind(workload_id).await?;
        self.workload_handles.write().await.remove(workload_id);
```

crates/wash-runtime/src/host/http.rs
```diff
@@ -43,11 +41,225 @@ use rustls_pemfile::{certs, private_key};
...
+        Ok(())
+    }
+
+    /// Pick a workload ID based on the incoming request
+    fn route_incoming_request(
+        &self,
+        req: &hyper::Request<hyper::body::Incoming>,
+    ) -> anyhow::Result<String> {
+        tokio::task::block_in_place(move || {
+            let lock = self.host_to_workload.blocking_lock();
```
ricochet: could use Arc<tokio::sync::RwLock> for better concurrent read performance:

```rust
  pub struct DynamicRouter {
      host_to_workload: Arc<tokio::sync::RwLock<HashMap<String, String>>>,
  }

  fn route_incoming_request(&self, req: &hyper::Request<..>) -> Result<String> {
      let lock = self.host_to_workload.try_read()
          .context("failed to acquire read lock")?;
      // ... rest of logic
  }
```

lxfontes: 👍 adding it!

crates/wash-runtime/src/host/http.rs
lxfontes: good call

ricochet:
Some more good opportunities for unit tests, but I think we should expand these out over time. The TODO's called out by this PR would be a good time to knock them out. I filed a ticket to make sure the ones I wanted to see are covered: https://github.com/wasmCloud/wasmCloud/issues/4945

crates/wash-runtime/src/host/http.rs
```diff
@@ -43,11 +41,225 @@ use rustls_pemfile::{certs, private_key};
...
+
+        let mut lock = self.host_to_workload.write().await;
+        lock.insert(host_header, resolved_handle.id().to_string());
+
+        Ok(())
+    }
+
+    async fn on_workload_unbind(&self, workload_id: &str) -> anyhow::Result<()> {
+        let mut lock = self.host_to_workload.write().await;
+        lock.remove(workload_id);
```
ricochet: I know this was mostly just moved, but pretty sure this is a bug now that the key can be a host header.

```suggestion
        lock.retain(|_host, wid| wid != workload_id);
```

Alternatively to catch these types of things going forward, we could type these values, e.g. `pub struct HostHeader(String); pub struct WorkloadId(String);`

```rust
host_to_workload: tokio::sync::Mutex<HashMap<HostHeader, WorkloadId>>
```

crates/wash-runtime/src/host/mod.rs
```diff
@@ -188,6 +190,7 @@ pub struct Host {
     /// System monitor for tracking CPU/memory usage
     system_monitor: Arc<RwLock<SystemMonitor>>,
     // endpoints: HashMap<String, EndpointConfiguration>
+    pub(crate) http_handler: std::sync::Arc<dyn crate::host::http::HostHandler>,
```
ricochet, line 193: This is just a suggestion, but I wonder if it would be more ergonomic to make this optional, e.g. `http_handler: Option<Arc<dyn HostHandler>>`. Then you can delete the NullServer.

lxfontes: I bounced between having/not-having the Option there.
The idea with NullServer is so embedders coming via the `wash_runtime::host` mod always has a valid http implementation to rely on and to provide the default behaviour for outgoing requests ( deny-all for now, but could be allow-all  too )
On the runtime side ( `wash_runtime::engine` ) it is indeed always optional ( `Ctx` struct ) and the default behaviour is always "deny-all".

not too strong about it tho, so if you feel this would better as `Option` I can change it.

ricochet: I think your thinking around it makes sense to me.

ricochet (approved):

PR #163, opened 2025-11-20
feat(wash): insecure option for oci commands

if0ne:


if0ne:
CC @ricochet

crates/wash/src/cli/oci.rs
```diff
@@ -52,13 +52,17 @@ pub struct PullCommand {
     /// The path to write the pulled component to
     #[clap(name = "component_path", default_value = "component.wasm")]
     component_path: PathBuf,
+    /// Use HTTP or HTTPS protocol
+    #[clap(long = "insecure", default_value_t = true)]
```
lxfontes: default to `false` ( as in: secure by default )

crates/wash/src/cli/oci.rs
```diff
@@ -90,6 +94,9 @@ pub struct PushCommand {
     /// The path to the component to push
     #[clap(name = "component_path")]
     component_path: PathBuf,
+    /// Use HTTP or HTTPS protocol
+    #[clap(long = "insecure", default_value_t = true)]
```
lxfontes: default to `false` ( as in: secure by default )

lxfontes (changes requested):
flip the default and 👍

if0ne:
done

lxfontes (approved):

PR #172, opened 2025-11-25
Remove wash new from templates

ricochet:
To reduce the scope for a MVP of wash v2.0, I propose we remove the templating feature until we can create, implement, and test them. Then we can build out a nice CLI experience based on language of choice, capabilities, etc. as an additive non-breaking feature.


lxfontes (approved):
agree 👍

PR #178, opened 2025-11-27
feat(wash/oci): check cached digest

if0ne:
## Feature or Problem

It might be worth making the check optional.


if0ne:
CC @lxfontes

lxfontes:
> It might be worth making the check optional.

💯 . This would be inline with Kubernetes `imagePullPolicy: Always | IfNotPresent`

I'm on the fence between removing artifact caching altogether & implementing imagePullPolicy.
rationale for removing being "less surprises".

lxfontes (approved):

PR #185, opened 2025-12-01
fix(runtime): init in dep order

### LINKED ISSUE #184: [BUG] Ordering error pre_instantiating components of a workload
Hi Guys,

We are encountering a problem with a multi-component Workload. One of the components, exports an interface that another component needs. However, linking of these components succeed +- 50% of the time. When it fails it does so with this linking error:

```
2025-12-01T13:19:17.493514Z  WARN wash_runtime::engine::workload: failed to link components, unbinding all plugins error=failed to pre-instantiate during component linking

Caused by:
    0: component imports instance `betty-blocks:data-api/data-api`, but a matching implementation was not found in the linker
    1: instance export `request` has the wrong type
    2: function implementation is missing
```

Did you encounter this before?

---
Steps to reproduce:

I am using a host build from the main branch, commit: c9d4eee00749a51245f0b7a503bc81ccc62705b2

1. Create three components
  A. Component that exports an incoming handler and imports am interface of B
  B. Component that exports interface called by A and imports interface of component C
  C. Component that exports interface called by B.
2. Start a host, with on only the `http_handler`
```
let http_port = find_available_port().await?;
    let http_addr: SocketAddr = format!("127.0.0.1:{http_port}").parse().unwrap();

    let engine = Engine::builder().build()?;

    let http_handler = DevRouter::default();
    let http_plugin = HttpServer::new(http_handler, http_addr);

    info!("Starting host");
    let host = HostBuilder::new()
        .with_engine(engine.clone())
        .with_http_handler(Arc::new(http_plugin))
        .build()?;

    let host = host.start().await?;
```
3. Add and start the workload:
```
let req = WorkloadStartRequest {
        workload_id: Uuid::new_v4().to_string(),
        workload: Workload {
            namespace: "test".to_string(),
            name: "test-workload".to_string(),
            annotations: HashMap::new(),
            service: None,
            components: vec![
                Component {
                    bytes: bytes::Bytes::from_static(COMPONENT_A),
                    local_resources: LocalResources {
                        memory_limit_mb: 256,
                        cpu_limit: 1,
                        config: HashMap::new(),
                        environment: config.clone(),
                        volume_mounts: vec![],
                        allowed_hosts: vec![],
                    },
                    pool_size: 1,
                    max_invocations: 100,
                },
                Component {
                    bytes: bytes::Bytes::from_static(COMPONENT_B),
                    local_resources: LocalResources {
                        memory_limit_mb: 256,
                        cpu_limit: 1,
                        config: HashMap::new(),
                        environment: config.clone(),
                        volume_mounts: vec![],
                        allowed_hosts: vec![],
                    },
                    pool_size: 1,
                    max_invocations: 100,
                },
                Component {
                    bytes: bytes::Bytes::from_static(COMPONENT_C),
                    local_resources: LocalResources {
                        memory_limit_mb: 256,
                        cpu_limit: 1,
                        config: HashMap::new(),
                        environment: config.clone(),
                        volume_mounts: vec![],
                        allowed_hosts: vec![],
                    },
                    pool_size: 1,
                    max_invocations: 100,
                },
            ],
            host_interfaces: vec![
                WitInterface {
                    namespace: "wasi".to_string(),
                    package: "http".to_string(),
                    interfaces: ["incoming-handler".to_string()].into_iter().collect(),
                    version: Some(semver::Version::parse("0.2.2").unwrap()),
                    config: {
                        let mut config = HashMap::new();
                        config.insert("host".to_string(), HOST_HEADER.to_string());
                        config
                    },
                },
                WitInterface {
                    namespace: "wasi".to_string(),
                    package: "http".to_string(),
                    interfaces: ["outgoing-handler".to_string()].into_iter().collect(),
                    version: Some(semver::Version::parse("0.2.4").unwrap()),
                    config: HashMap::new(),
                },
            ],
            volumes: vec![],
        },
    };

    info!("Starting workload");
    let request = test_config
        .host
        .workload_start(req)
        .await
        .context("Failed to start workload")?;
```

This step will fail approximately 50% of the time.

4. Call via HTTP request

This step succeeds the other 50% of the time.

---

I have looked into the host code, and the error seems to occur in the `pre_instantiate()` function in `workload.rs`. 
I am not sure how the internals of `wasmtime` work, but I have the feeling that when you export and import, the ordering of calling `instantiate_pre()` in `linker.rs` does matter a lot. First the component that exports should be added to `wasmtime` and after that the component that `imports`. A bit like is  written down here: https://github.com/bytecodealliance/wasmtime/blob/main/examples/linking.rs
However, currently, calling of `pre_instantiate()` feels random, and therefore it fails 50% of the time.

If you need more information, please let me know :)

Kind regards,
Mees

ricochet:
Fixes #184

### DIFF
--- a/crates/wash-runtime/src/engine/workload.rs
+++ b/crates/wash-runtime/src/engine/workload.rs
@@ -505,12 +505,50 @@ impl ResolvedWorkload {
 
     /// This function plugs a components imports with the exports of other components
     /// that are already loaded in the plugin system.
+    ///
+    /// Components are processed in topological order based on their inter-component
+    /// dependencies. This ensures that when a component imports from another component,
+    /// the exporting component has already had its imports resolved and can be
+    /// pre-instantiated.
     async fn resolve_workload_imports(
         &mut self,
         interface_map: &HashMap<String, Arc<str>>,
     ) -> anyhow::Result<()> {
-        let component_ids: Vec<Arc<str>> = self.components.read().await.keys().cloned().collect();
-        for component_id in component_ids {
+        // Build a dependency graph: for each component, track which other components it imports from
+        let mut dependencies: HashMap<Arc<str>, HashSet<Arc<str>>> = HashMap::new();
+
+        {
+            let components = self.components.read().await;
+            for (component_id, component) in components.iter() {
+                let mut deps = HashSet::new();
+                let ty = component.metadata.component.component_type();
+                for (import_name, import_item) in ty.imports(component.metadata.component.engine())
+                {
+                    if matches!(import_item, ComponentItem::ComponentInstance(_))
+                        && let Some(exporter_id) = interface_map.get(import_name)
+                        && exporter_id != component_id
+                    {
+                        // This import is provided by another component in the workload
+                        deps.insert(exporter_id.clone());
+                    }
+                }
+                dependencies.insert(component_id.clone(), deps);
+            }
+        }
+
+        // Topologically sort components: components with no dependencies (or dependencies
+        // already processed) come first. This ensures that when we process a component
+        // that imports from another component, the exporter has already been resolved.
+        let sorted_component_ids = topological_sort_components(&dependencies).context(
+            "failed to determine component processing order - possible circular dependency",
+        )?;
+
+        trace!(
+            order = ?sorted_component_ids.iter().map(|id| id.as_ref()).collect::<Vec<_>>(),
+            "processing components in topological order"
+        );
+
+        for component_id in sorted_component_ids {
             // In order to have mutable access to both the workload component and components that need
             // to be instantiated as "plugins" during linking, we remove and re-add the component to the list.
             let mut workload_component = {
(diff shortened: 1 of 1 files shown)


if0ne:
I believe cyclic imports are a valid scenario. 

Maybe it's necessary to implement a deferred call for creating the component instance? For example, call `instantiate` and cache it in `func_new_async`.

ricochet:
> I believe cyclic imports are a valid scenario.
> 
> Maybe it's necessary to implement a deferred call for creating the component instance? For example, call `instantiate` and cache it in `func_new_async`.

I took a bit to re-asses and review the spec. 

Some references that are handy:
1. Component definitions are acyclic - https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md?plain=1#L109
2. Acyclicity of module instantiation - https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md?plain=1#L347
3. WIT interfaces must be acyclic - https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md?plain=1#L431
4. No cross-component cycle collector - https://github.com/WebAssembly/component-model/blob/main/design/high-level/Choices.md?plain=1#L22
5. Breaking cyclic dependencies in shared-everything linking -
https://github.com/WebAssembly/component-model/blob/main/design/mvp/examples/SharedEverythingDynamicLinking.md?plain=1#L315

That last one is interesting where the SharedEverythingDynamicLinking.md document describes a workaround for cyclic dependencies within a composed component using indirect function calls. However, this is a specialized technique that:
- Applies to modules within a single component, not separate components
- Requires explicitly designed interfaces with call_indirect patterns
- Is specifically for the "shared-everything" linking model (shared memory, tables)

Since current code already defers instantiation - components are not instantiated until their functions are called (see lines 696-711 in resolve_component_imports). To link component A's imports to component B's exports, B must be pre-instantiatable first. This requires B's linker to be complete. I think this is the best we can do for now, at least until we get more CM features like runtime instantiation for parent components of children.

lxfontes (approved):

PR #193, opened 2025-12-08
feat(wash-runtime): outgoing params

if0ne:
## Feature or Problem

Right now, the limit that components can return through the HTTP handler is 2 megabytes. Sometimes components may exceed this limit, but in that case calling `wasi_http_incoming_handler().call_handle` freeze response thread. To fix this, you need to increase the outgoing parameters.


if0ne:
CC @lxfontes

lxfontes:
looking good! just need to get tests passing ✅

if0ne:
I created an issue in Wasmtime. Maybe this freeze behavior isn’t supposed to happen:

https://github.com/bytecodealliance/wasmtime/issues/12141

if0ne:
I believe that if we don’t need to expose these parameters for configuration, then the preferred solution to this issue should be #195

lxfontes:
addressed via #195 

lets wait for wasmtime 41 and revisit if we need to expose parameter 👍

PR #195, opened 2025-12-09
fix: call http handler in separate task

if0ne:
## Feature or Problem

The solution to the problem described here https://github.com/wasmCloud/wash/pull/193


if0ne:
CC @lxfontes

lxfontes (approved):

PR #209, opened 2025-12-18
chore: Multi-Wasm development refactor

lxfontes:
This PR brings multiple improvements to wash config, moving it from "Component development" to "Workload development". This better aligns the development & deployment stories.

Developers will be able to:
- Load extra Wasm components besides the component under development
- Mark the component under development as "Service" ( long running )
- Bring extra Wasm component as "Service"
- Enable / Disable WebGPU via project configuration
- Set the HTTP Server address via configuration
- Pass host directories to be mounted inside components via configuration

To see it in action:
`cargo run --  -C ./examples/cron-component dev`
or
`cargo run --  -C ./examples/cron-service dev`

### Removing "Dev Plugins"

This PR also removes the concept of "Dev Plugin".

Dev Plugins are Wash Plugins that have a `DevRegister` hook, bringing the entire plugin imports/exports into the same context as the component under development.

The same can be achieved by adding extra components via project config.

Note I'm not removing the `dev-register` entry from the WIT interface. this is because removing it would cause existing plugins ( ex: openapi2mcp ) to stop working.


crates/wash/src/config.rs
```diff
@@ -63,6 +69,57 @@ impl Config {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DevVolume {
+    // Host path to mount
+    pub host_path: PathBuf,
+    // Guest path inside the dev environment
+    pub guest_path: PathBuf,
+}
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct DevConfig {
```
lxfontes, line 99: This is `wash dev` section in the config file.

crates/wash/src/cli/dev.rs
```diff
@@ -495,179 +422,124 @@ impl CliCommand for DevCommand {
...
+        });
+    });
 
-        // Type-only interfaces that don't need plugin binding
-        // These are just type definitions used by other interfaces
-        if interface.interfaces.iter().any(|i| i == "types") {
-            return true;
+    // Extract both imports and exports from the component
+    // This populates host_interfaces which is checked bidirectionally during plugin binding
+    let mut host_interfaces = dev_config.host_interfaces.clone();
```
lxfontes, line 454: `host_interfaces` in the configuration takes precedence over interface inference, allowing users to pass configuration values to host plugins.

crates/wash/src/config.rs
```diff
@@ -63,6 +69,57 @@ impl Config {
...
+    /// Address for the dev server to bind to (default: "0.0.0:8000")
```
ricochet:
```suggestion
    /// Address for the dev server to bind to (default: "0.0.0.0:8000")
```

crates/wash/src/config.rs
```diff
@@ -63,6 +69,57 @@ impl Config {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address: Option<String>,
+
+    /// Whether the component under development should be treated as a service
+    #[serde(default)]
+    pub service: bool,
+    /// Optional path to a wasm component to be used as a service
+    pub service_path: Option<PathBuf>,
```
ricochet:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub service_path: Option<PathBuf>,
```

crates/wash/src/cli/dev.rs
```diff
@@ -495,179 +422,124 @@ impl CliCommand for DevCommand {
...
+                        bytes: Bytes::from(service_bytes),
+                        max_restarts: 0,
+                        local_resources: LocalResources {
+                            volume_mounts: volume_mounts.clone(),
+                            ..Default::default()
+                        },
+                    });
+                    debug!(path = ?dev_config.service_path.as_ref().unwrap().display(), "added service component to workload");
+                }
+                Err(e) => {
```
ricochet: nit: `std::io::Error` implements `Into<anyhow::Error>`. Consider using `?` operator with .context()


let service_bytes = std::fs::read(&dev_config.service_path)
      .with_context(|| format!("failed to read service component: {}", &dev_config.service_path.display()))?;

crates/wash/src/cli/dev.rs
```diff
@@ -495,179 +422,124 @@ impl CliCommand for DevCommand {
...
+                    },
+                    pool_size: -1,
+                    max_invocations: -1,
+                });
+                debug!(path = ?comp_path.display(), "added additional component to workload");
             }
-            // wasi:config/store gets the wasi config
-            ("wasi", "config") if interface.interfaces.contains("store") => {
-                interface.config = wasi_config.clone();
+            Err(e) => {
```
ricochet: Simplify here too: let comp_bytes = std::fs::read(comp_path)
      .with_context(|| format!("failed to read component: {}", comp_path.display()))?;

crates/wash/src/cli/dev.rs
```diff
@@ -239,35 +197,16 @@ impl CliCommand for DevCommand {
...
-
-        // Workload structure
-        let mut workload = create_workload(
-            wasm_bytes.into(),
-            wasi_config,
-            temp_mount_path,
-            dev_register_components,
-        );
+        // First run
+        let workload = create_workload(&host, &config, wasm_bytes.into())?;
```
ricochet: Right now we recreate the workload on every file change. IMO this is fine for now and avoids potential bugs with any kind of strategy we might use.

crates/wash/src/cli/dev.rs
```diff
@@ -159,37 +128,26 @@ impl CliCommand for DevCommand {
...
         // Empty context for pre-hooks, consider adding more
         ctx.call_hooks(HookType::BeforeDev, Arc::default()).await;
-        let dev_register_plugins = ctx.plugin_manager().get_hooks(HookType::DevRegister).await;
 
-        let mut dev_register_components = Vec::with_capacity(dev_register_plugins.len());
-        for plugin in dev_register_plugins {
-            dev_register_components.push(plugin.get_original_component(ctx).await?)
-        }
+        let dev_config = config.dev();
+        let http_addr = match &dev_config.address {
```
ricochet:
```suggestion
        let http_addr = dev_config.address.clone().unwrap_or_else(|| "0.0.0.0:8000".to_string());
```

crates/wash/src/config.rs
```diff
@@ -63,6 +69,57 @@ impl Config {
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DevVolume {
+    // Host path to mount
+    pub host_path: PathBuf,
+    // Guest path inside the dev environment
+    pub guest_path: PathBuf,
```
ricochet:
```suggestion
    /// Host path to mount
    pub host_path: PathBuf,
    /// Guest path inside the dev environment
    pub guest_path: PathBuf,
```

crates/wash/src/cli/dev.rs
ricochet: BTW shouldn't this be tokio::fs::read ?

crates/wash/src/config.rs
```diff
@@ -63,6 +69,57 @@ impl Config {
...
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+pub struct DevConfig {
+    /// Address for the dev server to bind to (default: "0.0.0:8000")
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub address: Option<String>,
+
+    /// Whether the component under development should be treated as a service
+    #[serde(default)]
+    pub service: bool,
```
ricochet, line 106: what happens if both service=true and service_path are specified? Do we handle that?

wit/world.wit
```diff
@@ -24,6 +24,7 @@ interface types {
         /// Called after `wash dev` ends the development loop
         after-dev,
         /// Called to register a component's exports for use in `wash dev`
+        /// This hook is deprecated and will be removed in future versions.
```
ricochet, line 27:
```suggestion
        /// This hook is deprecated and will be removed in future versions.
        /// @deprecated(version = 0.0.2)
```

crates/wash/src/config.rs
```diff
@@ -63,6 +69,57 @@ impl Config {
...
+    /// Optional path to a wasm component to be used as a service
+    pub service_path: Option<PathBuf>,
+
+    /// Additional components to load alongside the main component
+    #[serde(default)]
+    pub components: Vec<PathBuf>,
```
ricochet: Do we want these to be named? That would give us flexibility for later like binding config to a specific component. The main thought is that we want symmetry with ResolvedWorkload

```
pub struct ResolvedWorkload {
 ...
    /// All components in the workload. This is behind a `RwLock` to support mutable
    /// access to the component linkers.
    components: Arc<RwLock<HashMap<Arc<str>, WorkloadComponent>>>,
}
```

ricochet:
This is looking good. I ran the following to give it a whirl:

```bash
# Build the example components first
cd examples/cron-service
../../target/release/wash build

cd ../cron-component
../../target/release/wash build

# Option 1: Try developing the cron-component (with service as dependency)
cd ../cron-component
../../target/release/wash dev

# Option 2: Try developing the cron-service (as a service with component dependency)
cd ../cron-service
../../target/release/wash dev
```

crates/wash/src/config.rs
lxfontes: pretty sure was a miss cause they are named on the proto side.
the key point for being named is observability given workload ids & component ids are ephemeral on the host side.

by naming components our observability keys would look 👍 
workload  = workload name/namespace
workload service = workload name/namespace ( same as workload cause theres only 1 service in a workload )
workload component = workload name/namespace + component name

crates/wash/src/config.rs
lxfontes: `service: true` takes precedence and `service_path` is ignored.

if your intention was to use service = true, it will work.
if your intention was to use service_path, it will fail with "missing wasi:cli/run"

crates/wash/src/config.rs
```diff
@@ -63,6 +69,66 @@ impl Config {
...
+        match &self.dev {
+            Some(dev) => dev.clone(),
+            None => DevConfig::default(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct DevVolume {
+    // Host path to mount
```
ricochet: I think these should be doc comments still

ricochet (approved):

PR #219, opened 2026-01-02
feat: make workload_id in host the kubernetes UID

Mees-Molenaar:
## Feature or Problem

When redeploying a Workload, before the previous Workload is resolved, this Workload will not be removed from the host. So the Workload will exist forever.

When the Kubernetes operator sends a `WorkloadStart` request to a Host, it will only get a Workload ID assigned when the `workload_start` command is completely finished. Therefore, when you stop a Workload by redeploying, it will send a `WorkloadStop` request without an ID, because it does not have a Workload ID assigned yet. 

In this PR, I decided that we keep the Scheduler's (In this case the runtime-operator) UID. This way, the scheduler always knows which Workload is started and it always can be stopped by calling `WorkloadStop` request with that ID.

Another way of solving this, would be to wait in the Workloads finalizer for a Workload ID, or if the Workload deployment Errored for some reason. However, I preferred the implemented solution over this.

## Release Information
`next`

## Consumer Impact

When restarting the host without the correct runtime-operator, it will set empty workload_ids in for example the workload_to_host maps for http routing. Thus, the host should and runtime-operators should match.

### Manual Verification

1. Deploy a workload
2. Check with logs that the new workload that is being made uses the same ID as the Kubernetes Workload UID.


automation-wasmcloud:

automation-wasmcloud:

PR #220, opened 2026-01-04
fix: prevent cargo cache invalidation in wash-runtime build script

Aditya1404Sal:
## Feature or Problem
The build script was watching entire example directories and the fixtures
directory with `cargo:rerun-if-changed`, which included their target/
subdirectories.

Because the build script itself builds into those target/
directories, this was creating a circular dependency that was invalidating cargo's
cache on every build, causing a rerun everytime.

Changed to watch only the source inputs that should trigger rebuilds:
- The Individual Cargo.toml
- src/ directory
- wit/ directory

This will prevent the build script from triggering itself when it writes
build artifacts, fixing the cache-busting issue.

## Related Issues
wasmCloud/wasmCloud#4947

### DIFF
--- a/crates/wash-runtime/build.rs
+++ b/crates/wash-runtime/build.rs
@@ -66,8 +66,6 @@ fn build_fixtures_rust(workspace_dir: &Path) -> anyhow::Result<()> {
             continue;
         }
 
-        println!("cargo:warning=Building example: {}", name);
-
         // Build the example
         let status = Command::new("cargo")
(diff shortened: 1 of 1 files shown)


dman-os (approved):
Thanks, you didn't have to jump on this immediately on a Sun :)

Aditya1404Sal:
> Thanks, you didn't have to jump on this immediately on a Sun :)

No problem at all! Was already online and it only took a few minutes to sort out.

lxfontes (approved):

PR #225, opened 2026-01-07
feat: Blobby + Blobstore for wash dev

lxfontes:
Making blobstore available to wash dev ( in memory implementation ).

Brings blobby over from wasmcloud and gives it a UI ( thanks Bailey for the HTML / JS ).

Screenshot 2026-01-06 at 7 35 41 PM


crates/wash-runtime/src/plugin/wasi_blobstore.rs
```diff
@@ -425,14 +436,17 @@ impl bindings::wasi::blobstore::container::HostContainer for Ctx {
...
         let empty_map = HashMap::new();
-        let workload_storage = storage.get(&self.id).unwrap_or(&empty_map);
+        let workload_storage = storage
+            .get(&self.workload_id.to_string())
+            .unwrap_or(&empty_map);
 
         match workload_storage.get(container_name) {
             Some(container_data) => {
                 let objects: Vec<String> = container_data.objects.keys().cloned().collect();
+                debug!("objects {:?}", objects);
```
ricochet: probably fine to keep, but wasn't sure if this was just leftover from developing

crates/wash/src/cli/dev.rs
```diff
@@ -180,6 +180,7 @@ impl CliCommand for DevCommand {
                 .context("failed to create blobstore root directory")?;
         }
         debug!(path = ?volume_root.display(), "using blobstore root directory");
+        host_builder = host_builder.with_plugin(Arc::new(WasiBlobstore::new(None)))?;
```
ricochet, line 184: Instead of None, should this be volume_root?

examples/blobby/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+        error!("Failed to write data to blobstore: {e}");
+        return Err(format!("Failed to write data to blobstore: {e}"));
+    }
+
+    let copied_bytes = std::io::copy(&mut data, &mut stream).map_err(|e| e.to_string())?;
+    if copied_bytes != content_length {
+        Err(
+            "Expected to copy {content_length} bytes, but only copied {copied_bytes} bytes"
+                .to_string(),
+        )
```
ricochet: This will print `{content_length}` and not an interpolated string
```suggestion
        Err(
            "Expected to copy {content_length} bytes, but only copied {copied_bytes} bytes"
        ))
```

examples/http-blobstore/src/lib.rs
```diff
@@ -8,6 +8,7 @@ use wasmcloud_component::{
 
 struct Component;
 
+
```
ricochet: nit: I think this was a stray newline. Convention is 1 newline between top-level items. rust fmt doesn't enforce max number of blank lines by default.

crates/wash/src/cli/dev.rs
lxfontes: that option is actually "max object size" for in-memory objects. this blobstore root was previously used by the blobstore dev plugin.
 will remove the argument for now and bring it back ( likely as config and not arg ) when we transition from blobstore memory to filesystem.

ricochet (approved):

PR #228, opened 2026-01-12
chore: Removing Project Inference

lxfontes:
Makes so wash always respects the build command from configuration, without guessing how a component should be built.

Templates will carry the appropriate build command in their `.wash/config.yaml`

Added `go-hello-world` example which shows how to build using makefiles.

The tl;dr:

```
# Skip codegen during dev
dev:
  command: make build    <--- if present, overrides build command during wash dev

build:
  command: make bindgen build
  component_path: path/to/wasm  <--- becomes `$WASH_COMPONENT_PATH`. If NOT present, "<dir-name>.wasm" is used
```

When `wash build` is invoked, it will:
- Error if no build command is present
- Shell-exec the build command ( meaning env vars will be correctly replaced in build command )
  - passing original environment + `$WASH_COMPONENT_PATH`
- Check if there is a file present at `$WASH_COMPONENT_PATH`
  - Reload workload if file exists

For `wash dev` the flow is similar, but we use `dev.command` instead of `build.command` if present.


examples/go-hello-world/go.mod
```diff
@@ -0,0 +1,34 @@
+module go-hello-world
+
+go 1.24
+
+require (
+	go.bytecodealliance.org/cm v0.3.0
+	go.wasmcloud.dev/component v0.0.9
+)
+
+replace go.wasmcloud.dev/component => /Users/lucas/code/src/github.com/wasmCloud/go/component
```
lxfontes: this Go example could come in a separate PR.

had to override cause latest published version is erroring out
```
# go.wasmcloud.dev/component/gen/wasi/http/types
../../../../../../pkg/mod/go.wasmcloud.dev/component@v0.0.9/gen/wasi/http/types/types.wit.go:1308:27: cannot convert (bool)(cm.U32ToBool((uint32)(result0))) (value of type bool) to type "go.bytecodealliance.org/cm".BoolResult
```

good news is that `main` branch works.

lxfontes: fixed via https://github.com/wasmCloud/wasmcloud.com/pull/1054

lxfontes:
cc @ericgregory as this changes wash config & arguments

crates/wash/src/cli/component_build.rs
```diff
@@ -537,624 +250,47 @@ impl ComponentBuilder {
...
-                Command::new(command)
-                    .args(args)
-                    .current_dir(&self.project_path),
-                command,
-                &args_strings,
-            )
+        let args_strings = vec!["-c".to_string(), self.command.clone()];
+
+        info!(command = self.command,  component_path = ?component_path, "executing build command");
+        let mut cmd = Command::new("sh")
```
ricochet: We need to switch on windows here, e.g. cmd /c

crates/wash/src/cli/component_build.rs
```diff
@@ -537,624 +250,47 @@ impl ComponentBuilder {
...
-
-    /// Execute a custom build command, completely overriding standard build logic
-    async fn execute_custom_command(&self, custom_command: &[String]) -> anyhow::Result<PathBuf> {
-        let Some((command, args)) = custom_command.split_first() else {
-            bail!("custom command must not be empty");
-        };
-
-        info!(command = command, args = ?args, "executing custom build command");
+        // Attempt to canonicalize the component path
+        let component_path = component_path.canonicalize().unwrap_or(component_path);
```
ricochet: This will always return the or case on initial builds because canonicalize errors if the path doesn't exist. 

Move this down to after the build finishes

https://doc.rust-lang.org/std/fs/fn.canonicalize.html#errors

crates/wash/src/cli/component_build.rs
```diff
@@ -537,624 +250,47 @@ impl ComponentBuilder {
...
-                        );
-                        return Ok(path);
-                    }
-                }
-            }
+        if std::fs::exists(&component_path)? {
+            return Ok(component_path);
         }
 
         Err(anyhow::anyhow!(
```
ricochet: Error needs an update to:
```
build command completed successfully but component not found at: {}
```

crates/wash/src/config.rs
```diff
@@ -72,13 +71,30 @@ impl Config {
...
     /// Get the development configuration, defaulting to [DevConfig::default()] if not set
     pub fn dev(&self) -> DevConfig {
-        match &self.dev {
-            Some(dev) => dev.clone(),
-            None => DevConfig::default(),
-        }
+        self.dev.clone().unwrap_or_default()
+    }
+
+    pub fn build(&self) -> BuildConfig {
```
ricochet, line 77: Add an error/validation for not having a build command.
```suggestion
    pub fn build(&self) -> BuildConfig {
        let build = self.build.clone().unwrap_or_default();
        if build.command.is_empty() {
                bail!("build.command is required in .wash/config.yaml");
        }
        Ok(build)
    }
```

crates/wash/src/cli/component_build.rs
```diff
@@ -537,624 +250,47 @@ impl ComponentBuilder {
...
-
-            // Add WIT world - this is required for TinyGo builds when WIT is present
-            let Some(wit_world) = &tinygo_config.wit_world else {
-                // Generate project config to ensure .wash/config.yaml exists with placeholder
-                let mut config_with_placeholder = config.clone();
-                let component_path_relative = PathBuf::from("build/output.wasm");
+        let component_path = build_config
+            .component_path
+            .clone()
+            .unwrap_or(format!("{}.wasm", project_dir_name).into());
```
ricochet, line 298: I think we keep it as-is, but I'm very split. This default works well for go, but we're leaning into being explicit rather than guessing.

examples/go-hello-world/Makefile
```diff
@@ -0,0 +1,14 @@
+
+# Disable go.work from repository toplevel as this example is not part of the monorepo
+GOWORK = off
+export GOWORK
+
+all: build
+
+.PHONY: build
+build:
+	  tinygo build -target wasip2 -wit-package ./wit -wit-world http-server -o go-hello-world.wasm ./
```
ricochet: I get formatting warnings here. Makefiles require hard tabs and this line has mixed tab/spaces.

ricochet (changes requested):
Looking good. I recommend rebasing to one commit with a breaking change footer or with feat!(wash): remove project inference to signal the breaking change in the gh release notes.

crates/wash/src/config.rs
```diff
@@ -72,13 +71,30 @@ impl Config {
...
+        self.build.clone().unwrap_or_default()
     }
 }
 
+/// Build configuration for different language toolchains
```
ricochet: Super nit but probably want to drop the reference to different language toolchains:
```suggestion
/// Configuration for building WebAssembly components
///
/// # Example
///
/// ```yaml
/// build:
///   command: cargo build --target wasm32-wasip2 --release
///   component_path: target/wasm32-wasip2/release/my_component.wasm
/// ```
```

ricochet (approved):

PR #232, opened 2026-01-13
chore: Moving all plugins under wash_runtime::plugin

lxfontes:
We had 2 types of plugins living in different rust modules:
- in-memory / development plugins in `wash_runtime::plugin`
- nats / cluster plugins in `wash_runtime::washlet::plugins`

This PR moves plugins to `wash_runtime::plugin::<interface-name>::<concrete-implementation>`.

wasi_config is the odd one as the behaviour was slightly different. It was refactored so a single implementation can fulfill both behaviours, namely:
- in dev: config comes from WitInterface configuration
- in cluster: config comes from WitInterface configuration plus wasi:cli environment variables


crates/wash-runtime/tests/integration_http_counter.rs
```diff
@@ -22,8 +22,8 @@ use wash_runtime::{
         http::{DevRouter, HttpServer},
     },
     plugin::{
-        wasi_blobstore::WasiBlobstore, wasi_config::WasiConfig, wasi_keyvalue::WasiKeyvalue,
-        wasi_logging::WasiLogging,
+        wasi_blobstore::InMemoryBlobstore, wasi_config::DynamicConfig,
+        wasi_keyvalue::InMemoryKeyValue, wasi_logging::TracingLogging,
```
ricochet: TracingLogging is the odd one out here where the rest are {Feature}{Implementation} so could be `TracingLogger`

ricochet (approved):

PR #233, opened 2026-01-15
feat: Filesystem implementation for blobstore & keyvalue

lxfontes:
Selects in-memory vs filesystem based on wash config value.
See blobby & http-counter wash config for example usage.

### DIFF
--- a/crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
+++ b/crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
@@ -1,0 +1,824 @@
+use std::collections::HashSet;
+use std::path::{Path, PathBuf};
+use std::sync::Arc;
+
+use crate::engine::ctx::{ActiveCtx, SharedCtx, extract_active_ctx};
+use crate::engine::workload::WorkloadComponent;
+use crate::plugin::HostPlugin;
+use crate::plugin::WorkloadTracker;
+use crate::wit::{WitInterface, WitWorld};
+use anyhow::Context;
+use async_nats::jetstream::object_store::{self, List, Object, ObjectStore};
+use futures::StreamExt;
+use tokio::io::AsyncReadExt;
+use tokio::sync::RwLock;
+use wasmtime::component::Resource;
+use wasmtime_wasi::p2::pipe::{AsyncReadStream, AsyncWriteStream};
+use wasmtime_wasi::p2::{InputStream, OutputStream};
+
+const PLUGIN_BLOBSTORE_ID: &str = "wasi-blobstore";
+
+mod bindings {
+    wasmtime::component::bindgen!({
+        world: "blobstore",
+        imports: { default: async | trappable },
+        with: {
+            "wasi:io": ::wasmtime_wasi::p2::bindings::io,
+            "wasi:blobstore/container/container": crate::plugin::wasi_blobstore::filesystem::ContainerData,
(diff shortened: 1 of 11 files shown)


examples/http-counter/.wash/config.yaml
```diff
@@ -1,3 +1,7 @@
 build:
   command: cargo build --target wasm32-wasip2 --release
   component_path: target/wasm32-wasip2/release/http_counter.wasm
+
+dev:
+  wasi_blobstore_path: tmp/blobstore
+  wasi_keyvalue_path: tmp/keyvalue
```
lxfontes, line 7: here

crates/wash-runtime/src/plugin/wasi_keyvalue/filesystem.rs
```diff
@@ -0,0 +1,487 @@
+#![deny(clippy::all)]
+//! # WASI KeyValue Memory Plugin
+//!
+//! This module implements `wasi:keyvalue@0.2.0-draft` interfaces using
+//! NATS JetStream as the backend storage.
```
ricochet:
```suggestion
//! Filesystem as the backend storage.
```

crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
```diff
@@ -0,0 +1,704 @@
...
+            .remove_workload_with_cleanup(workload_id, workload_cleanup, component_cleanup)
+            .await;
+
+        Ok(())
+    }
+}
+
+fn list_files_recursively(path: impl AsRef<Path>, files: &mut Vec<PathBuf>) -> anyhow::Result<()> {
+    // Read the entries in the current directory
+    for entry_result in std::fs::read_dir(path)? {
```
ricochet, line 795: We could consider making this async with `tokio::fs::read_dir(path)` and `#[async_recursion]`. IMO If it becomes a problem, that's when we should reconsider the impl here

crates/wash-runtime/src/plugin/wasi_blobstore/filesystem.rs
```diff
@@ -0,0 +1,704 @@
...
+    async fn get_data(
+        &mut self,
+        container: Resource<ContainerData>,
+        name: ObjectName,
+        _start: u64,
+        _end: u64,
+    ) -> anyhow::Result<Result<Resource<IncomingValueHandle>, ContainerError>> {
+        let container_data = self.table.get(&container)?;
+
+        let file = container_data.root.join(name.as_str());
```
ricochet: We should validate this isn't reaching outside the container root. 

Maybe this?

```rust
  fn safe_path(root: &Path, untrusted: &str) -> Result<PathBuf, String> {
      let path = Path::new(untrusted);

      // Early rejection of obvious attacks
      if path.is_absolute() {
          return Err("absolute paths not allowed".into());
      }

      for component in path.components() {
          if matches!(component, std::path::Component::ParentDir) {
              return Err("parent directory references not allowed".into());
          }
      }

      let joined = root.join(path);

      // If file exists, verify canonical path
      if joined.exists() {
          let canonical = joined.canonicalize()
              .map_err(|e| format!("failed to resolve path: {e}"))?;
          if !canonical.starts_with(root) {
              return Err("path escapes container root".into());
          }
          return Ok(canonical);
      }

      Ok(joined)
  }
```

ricochet (approved):

PR #234, opened 2026-01-15
feat: qrcode example

lxfontes:
modernizing our example stack to use `wstd` instead of `wasmcloud_component`. Includes:
- How to handle path routing
- How to handle stack errors ( catch all / HTTP 500 )
- How to include assets in the component
- How to parse json input

Screenshot 2026-01-15 at 10 08 35 AM


examples/qrcode/wit/world.wit
```diff
@@ -0,0 +1,5 @@
+package wasmcloud:hello;
+
+world hello {
+   export wasi:http/incoming-handler@0.2.2;
```
ricochet, line 4: We should do a pass for normalizing all of our imports here. Next version of wasmtime should hav 0.2.9

examples/qrcode/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+        "/" => home(req).await,
+        "/qrcode" => qrcode(req).await,
+        _ => not_found(req).await,
+    }
+}
+
+async fn home(_req: Request<Body>) -> Result<Response<Body>, wstd::http::Error> {
+    Response::builder()
+        .status(StatusCode::OK)
+        .body(UI_HTML.into())
```
ricochet, line 36:
```suggestion
        .body(UI_HTML.into())
        .header("Content-Type", "text/html")
```

examples/qrcode/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+version = "0.1.0"
+
+[workspace]
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+wstd = "0.6.3"
+qirust = "0.1"
```
ricochet: we might want something that's been around longer with https://crates.io/crates/qrcode

ricochet (approved):

PR #236, opened 2026-01-16
chore: lint unwrap and expect

ricochet:
Summary:
- Added [lints] workspace = true to inherit workspace lints
- Added [workspace.lints.clippy] with unwrap_used = 'deny' and expect_used = 'deny'
- Allows in build.rs where it is valid to end execution on failure. 
- Allows in tests for convenience.
- crates/wasi/src/lib.rs - crate-level allows for unwrap_used and expect_used (wasmtime-wasi fork)                                                                                        

Code Changes          
```text                                                                                                                                                                          
┌────────────────────┬────────────────────────────────────┬──────────────────────────────────────────┐                                                                                        
│        File        │               Change               │                 Approach                 │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ update.rs          │ next().unwrap()                    │ → ok_or_else() (proper error)            │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ cli/wit.rs         │ file_name().unwrap() in debug logs │ → map().unwrap_or_default()              │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ plugin/mod.rs      │ strip_prefix().unwrap()            │ → unwrap_or() (simplified)               │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ plugin/runner.rs   │ table.get().unwrap()               │ → expect() + #[allow] (trait constraint) │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ main.rs            │ static directive parsing           │ → #[allow] (infallible)                  │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ main.rs            │ stdout write in exit               │ → #[allow] (exit context)                │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ http.rs            │ HTTP response building             │ → extracted error_response() helper      │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
│ workload.rs:833    │ func_new_async().expect()          │ → .context()? (proper error)             │                                                                                        
├────────────────────┼────────────────────────────────────┼──────────────────────────────────────────┤                                                                                        
[...]

Issue #235, opened 2026-01-15
Enforce additional clippy lints


ricochet:
Enforce additional clippy lints in `Cargo.toml` and have crates inherit.

```toml
[lints.clippy]
# Correctness & Safety
unwrap_used = "warn"          # Prefer expect() or proper error handling
expect_used = "warn"          # Even stricter - force Result handling
panic = "warn"                # No panics in library code
indexing_slicing = "warn"     # Use .get() instead of [i]

# Performance
inefficient_to_string = "warn"
needless_collect = "warn"
needless_pass_by_value = "warn"
trivially_copy_pass_by_ref = "warn"
unnecessary_box_returns = "warn"

# Code Quality
clone_on_ref_ptr = "warn"     # Explicit Arc::clone() over .clone()
dbg_macro = "warn"            # No dbg! in production
print_stdout = "warn"         # No println! in libraries
todo = "warn"                 # No TODO in production
```

`cargo clippy -- -W clippy::pedantic 2>&1 | grep -o -i warning | wc -l` came in at 735. So I do not recommend starting there.

ricochet:
After #246, the remaining ones require more changes or are more iffy if we even want these:
                                                                                           
  1. needless_pass_by_value - flagged idiomatic impl ToString patterns in                  
  CommandOutput::ok() and CommandOutput::error()                                           
  2. todo - there's a todo!("use common dir") in crates/wash/src/wit.rs:180 that needs     
  actual implementation                                                                    
  3. clone_on_ref_ptr - many violations, requires changing arc.clone() → Arc::clone(&arc)  
  throughout

if0ne:
Hey. It might also be worth adding `#[forbid(unsafe_code)]`

ricochet:
Closing this tracking issue now that the lints we want in have landed.

PR #237, opened 2026-01-16
warnings as errors

ricochet:
Merge after #236

This makes it so that our CI and our local clippy errors the same keeping our config in one source of truth (Cargo.toml). We do not want warnings to be introduced to the codebase from this point forward.


ricochet:
Strict enforcement for #235

lxfontes (approved):

PR #240, opened 2026-01-16
fix(rt): plugins binding for service

### LINKED ISSUE #188: If a workload's service attempts to use a plugin-provided interface, the entire workload will fail.
When a workload's service attempts to use a wit interface provided by a plugin, the host will print to it's stdout/err the following:
```
Error handling command: component not found during plugin binding
```

This appears to be caused by [this check](https://github.com/wasmCloud/wash/blob/main/crates/wash-runtime/src/engine/workload.rs#L1182) which will always fail if the requested interface is in a service, because the list which that is populated by, has [seperate handling](https://github.com/wasmCloud/wash/blob/main/crates/wash-runtime/src/engine/workload.rs#L1077) for the service part of the workload, and thus the service's ID won't be in `self.components`

I'm not sure,  but I suspect a similar problem exists when linking the components and the service, because if I hard-code everything my service needs to not use the plugins, my component is failing to start because the wit interface my service provides is missing:

```
2025-12-02T17:38:15.844037Z  WARN failed to link components, unbinding all plugins error=failed to pre-instantiate during component linking

Caused by:
    0: component imports instance `liizzii:ethermesh/ethermesh-send@0.1.0`, but a matching implementation was not found in the linker
    1: instance export `send` has the wrong type
    2: function implementation is missing
```

if0ne:
Closes #188 

I have two questions regarding the service and plugins:

1. Do we need to add `on_service_bind` for `HostPlugin`, or make an enum that is passed into `on_component_bind`, or should this never be called for a service?

2. Do we need validation for exported interfaces for a service to avoid confusing user. For example, they might export `wasmcloud:messaging` and expect the message to be delivered to a running service but it's not true. Or is it sufficient to cover this in the documentation?


if0ne:
CC @lxfontes @ricochet

ricochet:
For question 2 , yes we should add validation. A service is a wasi:cli/run component (has a main function that runs continuously). If a service exports wasmcloud:messaging/handler, those exports are never invoked. At least until/if we add support for long-running reactors.

if0ne:
I decided to add an enum and renamed `on_component_bind` to `on_workload_item_bind`. For the most part, these methods only use the id and the linker of the component or service, so I didn’t create a separate method.

examples/cpu-usage-service/.wash/config.yaml
```diff
@@ -0,0 +1,6 @@
+build:
+  command: cargo build --target wasm32-wasip2 --release
+  component_path: target/wasm32-wasip2/release/cron-service.wasm
```
lxfontes: `cpu-usage-service.wasm`

examples/messaging-handler/README.md
```diff
@@ -0,0 +1,19 @@
+# `http-hello-world`
```
lxfontes: 🔪  this readme or update it with a tl;dr

lxfontes (changes requested):
looking great !

lxfontes:
this PR is 🤏 close to done.
will hold off other PRs to avoid merge conflicts here.

if0ne:
done

lxfontes (approved):

PR #245, opened 2026-01-20
chore: Release Candidate 6

lxfontes:
- [x] Bump version in Cargo.toml
- [x] Bump version in runtime-operator chart
- [x] Sync CRD definition from runtime-operator to chart
- [x] Validate `runtime-operator/config/samples` work with no extra config
- [ ] ~Remove patch from Cargo.toml~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6347,7 +6347,7 @@ dependencies = [
 
 [[package]]
 name = "wash"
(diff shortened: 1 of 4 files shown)


ricochet (approved):
Removing the path should make zero functional difference. Probably worth doing this cleanup now.

PR #258, opened 2026-01-26
feat(rt): add error handling `workload_start`

if0ne:
In the current implementation `HostWorkload::Error` is unused. I think it should be used when `workload_start` fails. For now `workload_start` do not set `HostWorkload::Error` if workload initialization fails, and the workload status remains `HostWorkload::Starting` .


if0ne:
CC @lxfontes 

There are some inconsistency: `workload_status` returns error if the workload not found but `workload_stop` returns `WorkloadStatus::Unspecified` in the same case. Should we unify this behavior?

lxfontes:
agree on `workload_start`.

for `status` & `stop`: they should indeed have the same behaviour.
`stop` is the odd one here as we encode too much information in the result so the caller ( operator ) can reflect status upwards:
- error: something wrong stopping the workload ( operator action: retry )
- unspecified: the input workload_id was not found  ( operator action: ignore workload, continue )
- anything else: the actual workload state ( operator action: update )

we shouldn't be returning a sentinel ( `unspecified` ) to derive behaviour. I think we might be missing a [WorkloadState](https://github.com/wasmCloud/wash/blob/main/proto/wasmcloud/runtime/v2/workload.proto#L20) here,  namely `WORKLOAD_STATE_NOT_FOUND`

Both `status` & `stop` returning `WORKLOAD_STATE_NOT_FOUND` when the workload is not in the hashmap. This ensures callers can differentiate when a workload id is not there versus when the workload is in error state.

lxfontes (approved):

PR #263, opened 2026-01-27
feat: Opentelemetry Setup

lxfontes:
Sets up OTEL for logs / traces ( no metrics yet ).

OTEL configuration is driven by [SDK environment variables](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/).

We are tracking:
- Scheduling: Workload start / status / stop
- Wasm Entrypoints: http server and wasmcloud messaging
- Plugins lifecycle: bind / unbind operations

For Plugin Authors: Any `HostApi` interaction is observed by the runtime, so no need to instrument those unless you're enriching context.

An example setup would look like:

Start Aspire Dashboard ( Otel receiver ) with
```
docker run --rm -it \
    -p 18888:18888 -p 5318:18889 \
    --name aspire-dashboard \
    -e DOTNET_DASHBOARD_UNSECURED_ALLOW_ANONYMOUS="true" \
    -e ASPIRE_ALLOW_UNSECURED_TRANSPORT=true \
    mcr.microsoft.com/dotnet/aspire-dashboard:9.1
```

Configure Otel:
```
export OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:5318
export OTEL_EXPORTER_OTLP_PROTOCOL=grpc
```

Start wash:
```
cargo run --  -C ./examples/blobby dev
```

You should be able to see:
- Aspire on http://localhost:18888/
- Blobby on http://localhost:8000/

Screenshot 2026-01-27 at 3 07 08 PM

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -81,8 +81,8 @@ flate2 = { workspace = true, features = ["rust_backend"] }
 humantime = { workspace = true }
 opentelemetry = { workspace = true }
 opentelemetry_sdk = { workspace = true }
-opentelemetry-semantic-conventions = { workspace = true }
-opentelemetry-otlp = { workspace = true, features = ["grpc-tonic"] }
+opentelemetry-semantic-conventions = { workspace = true, features = ["semconv_experimental"] }
+opentelemetry-otlp = { workspace = true, features = ["grpc-tonic", "trace", "logs"] }
 opentelemetry-appender-tracing = { workspace = true }
 reqwest = { workspace = true, features = ["json", "rustls-tls"] }
 rustls = { workspace = true }
@@ -137,13 +137,13 @@ kube-derive = { version = "1", default-features = false }
 notify = { version = "8.0.0", default-features = false, features = ["macos_fsevent"] }
 oci-client = { version = "0.15.0", default-features = false, features = ["rustls-tls", "rustls-tls-native-roots"]}
 oci-wasm = { version = "0.3.0", default-features = false, features = ["rustls-tls"] }
-opentelemetry = { version = "0.28", default-features = false }
-opentelemetry-appender-tracing = { version = "0.28", default-features = false }
-opentelemetry-otlp = { version = "0.28", default-features = false }
-opentelemetry-http = { version = "0.28", default-features = false }
-opentelemetry_sdk = { version = "0.28", default-features = false }
-opentelemetry-semantic-conventions = { version = "0.28", default-features = false }
-opentelemetry-stdout = { version = "0.28.0", default-features = false }
+opentelemetry = { version = "0.31", default-features = false }
+opentelemetry-appender-tracing = { version = "0.31", default-features = false }
+opentelemetry-otlp = { version = "0.31", default-features = false }
+opentelemetry-http = { version = "0.31", default-features = false }
+opentelemetry_sdk = { version = "0.31", default-features = false }
+opentelemetry-semantic-conventions = { version = "0.31", default-features = false }
+opentelemetry-stdout = { version = "0.31.0", default-features = false }
 pbjson = { version = "0.8.0", default-features = false }
 pbjson-types = { version = "0.8.0", default-features = false }
 pbjson-build = { version = "0.8.0", default-features = false }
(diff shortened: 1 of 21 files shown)


src/main.rs
```diff
@@ -381,56 +404,103 @@ where
 /// Initialize tracing with a custom format
 ///
 /// Returns a tuple of stdout and stderr writers for consistency with the previous API.
-fn initialize_tracing(
+fn initialize_observability(
```
ricochet, line 406: OTLP export is always-on with no opt-out. Should we gate behind an exporter being set, e.g. `OTEL_EXPORTER_OTLP_ENDPOINT`?

src/main.rs
```diff
@@ -469,3 +539,12 @@ fn find_project_root() -> PathBuf {
 
     fallback
 }
+
+fn default_log_level() -> Level {
```
ricochet: RUST_LOG is typically a filter directive so I think this will end up biting folks. `Level::from_str("wash=debug")` will fail and silently fall back to INFO. I think we should plumb this one through as-is if it's defined.

Cargo.toml
```diff
@@ -164,6 +171,7 @@ tonic = { version = "0.14", default-features = false }
 tonic-prost = { version = "0.14", default-features = false }
 tonic-prost-build = { version = "0.14", default-features = false }
 tracing = { version = "0.1.41", default-features = false, features = ["attributes"] }
+tracing-opentelemetry = { version = "0.29", default-features = false }
```
ricochet: opentelemetry-otlp and related crates perhaps should go ahead and bump to 0.31 as of right now we have two versions of tonic coming in.

src/main.rs
```diff
@@ -381,56 +404,103 @@ where
...
-        // Register all layers with the subscriber
-        Registry::default().with(env_filter).with(fmt_layer).init();
-
-        (Box::new(std::io::stdout()), Box::new(std::io::stderr()))
+            .add_directive(directive("wasm_pkg_client=error")?)
+            .add_directive(directive("wasm_pkg_core=error")?);
     }
+
+    let fmt_layer = tracing_subscriber::fmt::layer()
+        .with_writer(std::io::stdout)
```
ricochet: I can be down with us always writing logs to stdout, but note this is a change from stderr to stdout. Some folks feel strongly about keeping logs separate from data, but I think that's really just preference when in practice places like container runtimes treat them identically.

src/main.rs
lxfontes: it's a ⚖️ against the rust-ism "RUST_LOG controls opentelemetry output".
The middle ground would be: For otel logs & traces we use a envfilter with default level from `log_level`, and for stdout we use a envfilter from RUST_LOG.

src/main.rs
lxfontes: will move it back to stderr.

src/main.rs
ricochet: I like that. It's another point in the keeping stdout and stderr split

if0ne:
I've been waiting for this. Could you add support for passing metadata map values from cli/env? https://docs.rs/tonic/latest/tonic/metadata/struct.MetadataMap.html

I want to use OpenObserve for observability, and it requires basic authentication headers.

src/main.rs
lxfontes: checking if any "OTEL_" exists 👍

Cargo.toml
lxfontes: bumped the whole group

lxfontes:
@if0ne metadata as in http headers or host/workload attributes?

for headers: the current code will respect https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_headers  so basic  auth is `export OTEL_EXPORTER_OTLP_HEADERS="Authorization: Bearer xxxx"

for metadata, was thinking of surfacing all or a subset of workloads' `annotations` as they can be enriched at many layers ( kube crd, operator itself, host ).

if0ne:
> @if0ne metadata as in http headers or host/workload attributes?
> 
> 
> 
> for headers: the current code will respect https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_headers  so basic  auth is `export OTEL_EXPORTER_OTLP_HEADERS="Authorization: Bearer xxxx"
> 
> 
> 
> for metadata, was thinking of surfacing all or a subset of workloads' `annotations` as they can be enriched at many layers ( kube crd, operator itself, host ).

As http headers. Thanks, I didn’t notice this env.

ricochet (approved):

PR #269, opened 2026-01-29
feat: Artifact & Compilation Caching

lxfontes:
While two distinct features, shipping them together due to their interaction.

# Artifact Caching

Follows [kubernetes imagepullpolicy expectations](https://kubernetes.io/docs/concepts/containers/images/#image-pull-policy)

There is a significant blast radius:
- host
- operator
- operator CRDs
- protobuf ( NATS api )

## Key changes

`host` mod:
- The `Host` now accepts an optional `cache_dir`. Embeders wishing to cleanup the cache can call `oci::cleanup_cache` at their desired interval.

`oci` mod:
- Introduced `OciPullPolicy`

`washlet` mod:
- Host Cache cleanup is always on (as cache is always on)
- Both Component & Service digests will be sourced from OCI

Operator CRDs:
- an optional `imagePullPolicy` is now available for Components & Services

# Compilation caching

Ability to cache compiled component, regardless if it came from OCI or not.
This alleviates one of the most expensive steps in workload initiation.

`types` mod:
- Both `Component` and `Service` have a `digest: Option<String>`, opt-in cache compilation at the component level

`engine` mod:
- `EngineBuilder` configures 
- `load_component_bytes` handles cache interaction

`washlet` mod:
- Both Component & Service will pass their OCI digest as compilation cache digest

related to wasmCloud/wasmCloud#4940


crates/wash-runtime/src/washlet/mod.rs
```diff
@@ -150,13 +150,22 @@ pub async fn run_cluster_host(
             .context("failed to subscribe for API requests")?;
         let mut heartbeat_timer = tokio::time::interval(heartbeat_interval);
 
+        let mut oci_cleanup_timer = tokio::time::interval(Duration::from_secs(300));
```
ricochet: We probably want to expose these as config including:

  - Cache cleanup interval:                    
  - Cache expiration age                                 
  - Compilation cache size                                                  
  - Compilation cache TTL

lxfontes: those are at a different level (hostbuilder), the only non-configurable knob atm is this `cleanup_timer` will add it to washlet.

the confusion here is real as we have 3 builders stacking values ( washlet / host / engine )

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -268,6 +286,37 @@ impl Engine {
...
+                let inner = &self.inner;
+                let bytes_ref = bytes.as_ref();
+
+                self.cache
+                    .try_get_with(key, || {
+                        Component::new(inner, bytes_ref)
+                            .context("failed to compile component from bytes")
+                            .map(CacheValue)
+                    })
+                    .map_err(|e| anyhow::anyhow!("{}", e))
```
ricochet: This stringifies the Arc'd error, losing the error chain.
```suggestion
                    .map_err(|e| anyhow::anyhow!(e).context("compilation cache error"))
```

ricochet (approved):

crates/wash-runtime/src/washlet/mod.rs
lxfontes: added to builder 👍

PR #270, opened 2026-01-29
feat: fine grained wasmtime configuration

lxfontes:
Allows wasmtime pooling allocator to be configured via environment variables.
The `1000` max_instances number is a guideline and can be configured while building a `Host`.

(maybe) related to wasmCloud/wasmCloud#4940


crates/wash-runtime/src/engine/mod.rs
```diff
@@ -406,6 +409,13 @@ impl EngineBuilder {
         self
     }
 
+    /// Sets the maximum number of instances for the pooling allocator.
+    /// This is a 'hint' and can be overidden by environment variables.
```
ricochet:
```suggestion
    /// This is a 'hint' and can be overridden by environment variables.
```

crates/wash-runtime/src/engine/mod.rs
```diff
@@ -442,18 +452,25 @@ impl EngineBuilder {
...
-        self.config.async_support(true);
-        // The pooling allocator can be more efficient for workloads with many short-lived instances
-        if let Ok(true) = use_pooling_allocator_by_default(self.use_pooling_allocator) {
-            tracing::debug!("using pooling allocator by default");
-            self.config
-                .allocation_strategy(wasmtime::InstanceAllocationStrategy::Pooling(
-                    PoolingAllocationConfig::default(),
+        // If a custom config was provided, use it as-is
+        let config = if let Some(cfg) = self.config.take() {
+            cfg
```
ricochet, line 462: Unused max_instances when custom config is provided 
```suggestion
            if self.max_instances.is_some() || self.use_pooling_allocator.is_some() {                    
              bail!("cannot use with_config() together with with_max_instances() or with_pooling_allocator()");       
            }

            cfg
```

Or alternatively use a typestate pattern where with_config() returns a different builder type that doesn't have  with_max_instances() or with_pooling_allocator() methods.

crates/wash-runtime/Cargo.toml
```diff
@@ -41,7 +41,7 @@ uuid = { workspace = true, features = ["v4"] }
 tempfile = { workspace = true }
 tokio = { workspace = true, features = ["sync", "net", "macros"] }
 tracing = { workspace = true }
-wasmtime = { workspace = true, features = ["component-model", "cranelift", "pooling-allocator"] }
+wasmtime = { workspace = true, features = ["component-model", "cranelift", "pooling-allocator", "gc", "gc-drc", "gc-null"] }
```
ricochet: I'd like for these to be behind our own experimental features flags. So we can use `x-wasmtime-gc` or `experimental-wasmtime-gc` for any experimental features we are enabling. This will need to be a top-level flag for wash-runtime. Maybe we do this as a follow-on to after the workspace refactor we want to make?

ricochet (approved):

crates/wash-runtime/Cargo.toml
lxfontes: trying to gate the gc config on `#[cfg(not(windows))]` so we can rollback this cargo toml change

PR #272, opened 2026-01-30
chore: Reorganizing Cargo Workspace

lxfontes:
Gives `wash` crate its own Cargo.toml.
Moves `main.rs` and `tests` over to wash crate.
Sets `wash` crate as the default workspace crate, so build commands remain the same.


crates/wash/Cargo.toml
```diff
@@ -0,0 +1,94 @@
+[package]
+name = "wash"
+categories = ["wasm", "command-line-utilities"]
+description = "The Wasm Shell (wash) for developing and publishing Wasm components"
+keywords = ["webassembly", "wasm", "component", "wash", "cli"]
+readme = "README.md"
```
ricochet: references readme but there's no README in this directory now

ricochet (approved):

crates/wash/Cargo.toml
lxfontes: got a minimal readme in there 👍

ricochet (approved):

PR #276, opened 2026-02-02
chore: prepare rc7

lxfontes:
🚀


Cargo.toml
```diff
@@ -11,7 +11,7 @@ members = [
 authors = ["The wasmCloud Team"]
 edition = "2024"
 rust-version = "1.91.0"
-version = "2.0.0-rc.6"
+version = "2.0.0-rc.7"
```
ricochet, line 14: need to run cargo check to get an updated lock

ricochet (approved):

PR #278, opened 2026-02-04
Add wasi-otel wash-runtime plugin

jfleitz:
## Add `wasi:otel` host plugin to `wash-runtime`

This PR adds first-class OpenTelemetry support for WebAssembly components running in `wash-runtime` by implementing the [`wasi:otel@0.2.0-rc.1`](https://github.com/WebAssembly/wasi-otel) interfaces. Components can now emit traces, structured logs, and metrics through standardized WASI APIs, with the host transparently exporting telemetry via OTLP/gRPC.

### What's new

**`wasi_otel` host plugin** — a new `HostPlugin` implementation in `crates/wash-runtime/src/plugin/wasi_otel/` (~840 lines across two files):

| Interface | Functions | Description |
|---|---|---|
| `wasi:otel/tracing` | `on-start`, `on-end`, `outer-span-context` | Span lifecycle management. Components create spans, attach attributes/events/links, and retrieve the host's current span context for trace propagation. |
| `wasi:otel/logs` | `on-emit` | Structured log emission. Log records carry severity, body, resource/scope metadata, and trace correlation IDs. |
| `wasi:otel/metrics` | `export` | Metrics export. Supports gauges, monotonic/non-monotonic sums, and histograms with resource and scope attribution. |
| `wasi:otel/types` | _(type definitions)_ | Shared types: `KeyValue`, `Resource`, `InstrumentationScope`, and all metric/trace/log record structures. |

The plugin manages three independent OpenTelemetry SDK providers (traces, logs, metrics), each backed by a batch OTLP/gRPC exporter. It handles per-component context tracking for span correlation and provides graceful shutdown with provider flushing.

**Type conversion layer** (`convert.rs`) — bidirectional conversion between WIT-defined `wasi:otel` types and the OpenTelemetry Rust SDK types, covering span contexts, log records, metric data points, severity levels, span kinds, and status codes.

### Configuration

The plugin is gated behind the `wasi-otel` cargo feature (enabled by default) and activated at runtime via the `wasi_otel` dev config option:

```yaml
# .wash/config.yaml
dev:
  wasi_otel: true
```

Standard OpenTelemetry environment variables are respected by the underlying SDK:
- `OTEL_EXPORTER_OTLP_ENDPOINT` — collector endpoint (default: `http://localhost:4317`)
- `OTEL_EXPORTER_OTLP_PROTOCOL` — transport protocol
- `OTEL_SERVICE_NAME` — service identity for exported telemetry

### Example: `otel-http`

A new `examples/otel-http/` project demonstrates all three observability pillars in a single Wasm component. The component implements an HTTP counter service that:

1. **Retrieves the host span context** via `outer-span-context` for distributed trace propagation
2. **Creates a parent span** (`handle-request`) with child spans for each downstream operation
3. **Makes an outgoing HTTP request** to `example.com` (child span with `http.method`, `http.url` attributes)
4. **Stores the response** in a blobstore container (child span with `blobstore.container` attributes)
5. **Increments a counter** in a key-value store (child span with `keyvalue.key` attributes)
6. **Emits structured logs** at each step, correlated with the active trace via `trace-id` and `span-id`
7. **Exports metrics** — a monotonic `U64Sum` counter for total requests and a `U64Gauge` for response body size

Run it with:
```bash
cd examples/otel-http
wash dev
```

Then send requests and observe telemetry flowing to your configured OTLP collector.

### Files changed

- **`crates/wash-runtime/src/plugin/wasi_otel/mod.rs`** — Core plugin: initialization, OTLP exporter setup, `HostPlugin` trait impl, and host-side implementations of all four `wasi:otel` interfaces
- **`crates/wash-runtime/src/plugin/wasi_otel/convert.rs`** — Type conversions between WIT and OpenTelemetry SDK types
- **`crates/wash-runtime/wit/deps/wasi-otel-0.2.0-rc.1/package.wit`** — WIT interface definitions for `wasi:otel@0.2.0-rc.1`
- **`crates/wash-runtime/wit/world.wit`** — Updated world to import `wasi:otel` interfaces
- **`crates/wash-runtime/Cargo.toml`** — New `wasi-otel` feature and dependencies (`opentelemetry-otlp`, `hex`)
- **`crates/wash/src/cli/dev.rs`** — Plugin registration in dev mode when `wasi_otel` is enabled
- **`crates/wash/src/config.rs`** — `wasi_otel` field added to `DevConfig`
- **`examples/otel-http/`** — Full example component demonstrating tracing, logging, and metrics

### Test plan

- [x] `cargo build` succeeds with default features
- [x] `cargo build --no-default-features` succeeds without the `wasi-otel` feature
- [x] Run `otel-http` example with `wash dev` and verify telemetry reaches an OTLP collector (e.g. Jaeger, Grafana Alloy, or `otel-desktop-viewer`)
- [x] Verify trace hierarchy: parent `handle-request` span with three child spans
- [x] Verify log records are correlated with traces via `trace-id` / `span-id`
- [x] Verify `http.server.request_count` counter and `http.server.response_body.size` gauge are exported
- [x] Test with `OTEL_EXPORTER_OTLP_ENDPOINT` override pointing to a remote collector

Resolves ENG-5576


if0ne:
Hey @lxfontes Maybe we should to move `initialize_observability` from wash-cli to wash-runtime so that we can reuse it to initialize observability when running the host and for the otel plugin?

lxfontes:
@if0ne I really would like to and it's ok to move / refactor / extract-parts of `initialize_observability` in this PR.


When plumbing OTEL, the challenges were:
- tracing subscriber registry manipulation: happens very early on, and disconnected from where `Host` is created
- how early to initialize OTEL: the early, the better
- if otel is a `host` or `engine` mod concern: `host` mod has start/stop lifecycles, `engine` doesn't

I think the only real blocker is tracing subscriber registry manipulation or some alternative to it (ex: separate tracing subscriber from otel). In wasmcloud v1 otel was split between [tracing](https://github.com/wasmCloud/wasmCloud/tree/250f4b2f6fe11f8814c872a2b8b41a244fc034bf/crates/tracing) and [core](https://github.com/wasmCloud/wasmCloud/blob/250f4b2f6fe11f8814c872a2b8b41a244fc034bf/crates/core/src/otel.rs) crates, which ⚖️  if might help or make it more complex. 


We don't need to address these here, mostly for context.

examples/otel-http/.wash/config.yaml
```diff
@@ -0,0 +1,8 @@
+build:
+  command: cargo build --target wasm32-wasip2 --release
+  component_path: target/wasm32-wasip2/release/http_counter.wasm
```
lxfontes: should be `otel_http`

examples/otel-http/.gitignore
```diff
@@ -0,0 +1,6 @@
+# Rust build artifacts
+target/
+
+# Wash build artifacts
+build/
+
```
lxfontes, line 6: add `wit/deps/`

Cargo.toml
```diff
@@ -72,7 +72,7 @@ opentelemetry = { version = "0.31", default-features = false }
 opentelemetry-appender-tracing = { version = "0.31", default-features = false }
 opentelemetry-otlp = { version = "0.31", default-features = false }
 opentelemetry-http = { version = "0.31", default-features = false }
-opentelemetry_sdk = { version = "0.31", default-features = false }
+opentelemetry_sdk = { version = "0.31", features = ["logs"] }
```
lxfontes: do we need `features` here given we have it in `wash-runtime` cargo toml ?

jfleitz: Good call. After re-basing (was branched from main prior to your cleanup/moving of things), I auto-merged the conflict that I had here. reverting back to just default-features=false

crates/wash-runtime/src/plugin/wasi_otel/mod.rs
```diff
@@ -0,0 +1,451 @@
...
+    async fn on_start(&mut self, span_context: bindings::wasi::otel::tracing::SpanContext) -> wasmtime::Result<()> {
+        // Log the span start - the actual span is managed by the guest
+        tracing::info!(
+            trace_id = %span_context.trace_id,
+            span_id = %span_context.span_id,
+            is_remote = span_context.is_remote,
+            "WASI span started"
+        );
+        Ok(())
+    }
```
EwoutV: Thanks @jfleitz for your work on `wasi:otel` in wasmCloud! I am currently experimenting with (automatic) OTEL instrumentation of statically composed components and ran into one issue with the current implementation of `wasi:otel`. If I understand correctly, `outer-span-context()` currently always returns the first span context created by the host. e.g. when the workload is called over HTTP, RPC, or a custom workload service. It is expected from the developer to keep track of span contexts within the component in order to create child spans for spans created within that component. This seems to be an important limitation for statically composed components with (automatic) OTEL instrumentation. Say a span is created in one component (call it `gateway.wasm`), it calls a function in another component (`service.wasm`), and this external function then creates a new span. When we `wac` the two components together, the second `service.wasm` component has no idea that it was called within a parent span, so context propagation breaks. A simple solution would be to make the span propagation part of the function's signature, but this seems rather suboptimal and also makes auto-instrumentation way harder.

I noticed that the current `on-start` implementation only performs some logging. Would it make sense for `on-start` to register the current span context whenever it is invoked? That could address the logical component-boundary propagation issue: `outer-span-context` would then return the context of the span that is currently active, rather than being permanently tied to the host’s root span. But please let me know if this would violate the proposed `wasi:otel` API or if I'm missing something!

jfleitz: Hello @EwoutV . The component's context is initialized with [OnBind](https://github.com/jfleitz/wash/blob/8c9f729eb0859c4e8f37fdd369ccaab140a423b6/crates/wash-runtime/src/plugin/wasi_otel/mod.rs#L206-L218). When you wac two components together, OnBind will be called once, as well as OnStart and OnEnd. So any tracing calls from the components will be within the component's context, not the Host context. Hope that makes sense!

EwoutV: Hi @jfleitz, thanks for getting back to me. So I'm not sure I'm following entirely. The issue I'm facing is that the current span context during a component's execution should be dynamic. For example, for a statically composed component being the result of having `wac`ed two original components A and B: when you create a span in component A, in which you invoke (a method of) component B, the span parent for component B should be the span created in component A, so the trace *should* look something like this:

```
Host span
 └── A span
      └── B span
```

With the current implementation (I think this is a limitation of the current `wasi:otel` proposal), component B doesn't know it is being executed in the span created by component A, since `outer-span-context` always returns the outer span created in `OnBind`. So the actual trace will incorrectly look like this, where each span created in a component will be a sibling of each other:

```
Host span
 ├── A span
 ├── B span
 ```

Maybe I'm still missing something, but I haven't been able to solve this issue with the current `wasi:otel` proposal and draft implementations.

lxfontes (approved):

lxfontes (approved):

PR #279, opened 2026-02-04
feat: wasmcloud-messaging template & in-memory implementation

lxfontes:
Screenshot 2026-02-04 at 6 31 28 PM

Getting the PR up for discussion


crates/wash-runtime/src/engine/mod.rs
```diff
@@ -464,6 +464,8 @@ impl EngineBuilder {
             let mut cfg = wasmtime::Config::default();
             // Async support must be enabled
             cfg.async_support(true);
+            cfg.wasm_component_model(true);
+            cfg.wasm_component_model_async(true);
```
lxfontes: I _think_ something within wstd is requiring this to be enabled. weird.

without it, the component cannot be loaded due to missing `subtask.cancel`

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+        }
+    }
+
+    #[instrument(skip_all, fields(subject = %msg.subject, reply_to = %msg.reply_to.as_deref().unwrap_or("<none>")))]
+    async fn publish(&mut self, msg: types::BrokerMessage) -> anyhow::Result<Result<(), String>> {
+        let Some(plugin) = self.get_plugin::<InMemoryMessaging>(PLUGIN_MESSAGING_MEMORY_ID) else {
+            return Ok(Err("plugin not available".to_string()));
+        };
+
+        let workload_id = self.ctx.workload_id.to_string();
```
lxfontes, line 167: This behaviour is strictly for single-workloads ( wash dev ).

In `wash dev`, we have a single workload so all subscriptions resolve to the same component.

In `wash host`, we use NATS for routing messages to multiple components.

The use case we would not be covering here is embedded ( as in: using the runtime directly ). One would expect to have an in-memory implementation that can communicate cross-workload. Like NATS, but all local.
We are not implementing that initially and will wait for feedback / requests.

templates/http-api-with-distributed-workloads/http-api/ui.html
```diff
@@ -0,0 +1,148 @@
...
+            background: #f5fff5;
+        }
+    </style>
+</head>
+<body>
+    <h1>Task API</h1>
+    <form id="taskForm">
+        <label for="worker">Worker</label>
+        <select id="worker" name="worker">
+            <option value="leet_speak" selected>Leet Speak</option>
```
ricochet:
```suggestion
            <option value="task-leet" selected>Leet Speak</option>
```

ricochet: This works today because all messages go to the same VecDeque regardless of subject, but doesn't hurt to make sure the component that subs on identity as `task-leet` matches.

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+            .context("failed to instantiate messaging pre")?;
+
+        let workload = workload.clone();
+        let component_id = component_id.to_string();
+
+        debug!("Spawning messaging processor for component {component_id}");
+
+        // Spawn the message processing task
+        tokio::spawn(async move {
+            let poll_interval = std::time::Duration::from_millis(10);
```
ricochet: for dev, probably fine. Could change to `tokio::sync::Notify`

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+        let pre = bindings::MessagingPre::new(instance_pre)
+            .context("failed to instantiate messaging pre")?;
+
+        let workload = workload.clone();
+        let component_id = component_id.to_string();
+
+        debug!("Spawning messaging processor for component {component_id}");
+
+        // Spawn the message processing task
+        tokio::spawn(async move {
```
ricochet: Spawned task is not tracked. The cancellation token handles graceful shutdown, but if the task panics, it will go unnoticed.

Either store the handle and track on `ComponentData`. 

```suggestion
        let task_component_id = component_id.clone();
        // Spawn the message processing task
        let handle = tokio::spawn(async move {
```

Or use tokio::spawn with .abort() on cleanup

templates/http-api-with-distributed-workloads/wit/world.wit
```diff
@@ -0,0 +1,10 @@
+package wasmcloud:template@0.1.0;
+
+world http-api {
```
ricochet, line 5:
```suggestion
world http-api {
  export wasi:http/incoming-handler@0.2.0;
```

templates/http-api-with-distributed-workloads/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+package.authors = ["The wasmCloud Team"]
+members = ["http-api", "task-leet"]
+
+[workspace.dependencies]
+wit-bindgen = "0.46.0"
+wstd = "0.6.3"
+serde = { version = "1.0", features = ["derive"] }
+anyhow = "1.0"
+async-trait = "0.1"
+uuid = { version = "1", features = ["v4"] }
```
ricochet: uuid is not used in this template. Also can async-trait be removed too?

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+
+        // Create the request message with reply_to set
+        let msg = types::BrokerMessage {
+            subject,
+            reply_to: Some(reply_to.clone()),
+            body,
+        };
+
+        debug!(subject = %msg.subject, reply_to = %msg.reply_to.as_deref().unwrap_or("<none>"), "Sending request");
+        // Push the message to the queue
```
ricochet, line 121: Avoid potential deadlock by avoiding grabbing pending_requests while still holding pending_messages.

```suggestion
        // Push the message to the queue
        let queue_full = {
            let mut msg_lock = pending_messages.write().await;

            if msg_lock.len() >= MAX_QUEUE_SIZE {
                true
            } else {
                msg_lock.push_back(msg);
                false
            }
        };

        if queue_full {
            let mut req_lock = pending_requests.write().await;
            req_lock.remove(&reply_to);
            return Ok(Err("message queue full".to_string()));
        }
```

templates/http-api-with-distributed-workloads/http-api/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+    let request_timeout = Duration::from_secs(5).as_millis() as u32;
+
+    let resp = consumer::request(subject, body, request_timeout)
+        .await
+        .map_err(anyhow::Error::msg)?;
+
+    Response::builder()
+        .status(StatusCode::OK)
+        .body(resp.body.into())
+        .map_err(Into::into)
```
ricochet: The current code uses .map_err(anyhow::Error::msg)? which converts the String error into an anyhow error that propagates up and likely produces an opaque 500. 

Could change it to return 502's for messaging errors (timeout, queue full, channel closed), and include the error string in the body:

```suggestion
    match consumer::request(subject, body, request_timeout).await {
        Ok(resp) => Response::builder()
            .status(StatusCode::OK)
            .body(resp.body.into())
            .map_err(Into::into),
        Err(err) => Response::builder()
            .status(StatusCode::BAD_GATEWAY)
            .body(err.into())
            .map_err(Into::into),
    }
```

crates/wash-runtime/src/plugin/wasmcloud_messaging/in_memory.rs
```diff
@@ -0,0 +1,372 @@
...
+use tracing::{Instrument, debug, instrument, warn};
+
+const PLUGIN_MESSAGING_MEMORY_ID: &str = "wasmcloud-messaging-memory";
+const MAX_QUEUE_SIZE: usize = 10000;
+
+mod bindings {
+    crate::wasmtime::component::bindgen!({
+        world: "messaging",
+        imports: { default: async | trappable | tracing },
+        exports: { default: async | tracing },
```
ricochet, line 19: AHA! This is why cm async was needed. Drop these?

ricochet: Ok, I dug in more.  The core issue is that wasmtime's async_support(true) (which the host needs) requires all component export calls to use call_async. The exports: { default: async } in the host-side bindgen generates this.

We can remove is async: true from the guest components' wit_bindgen::generate!. That's what was producing the subtask.cancel wasm instruction.

templates/http-api-with-distributed-workloads/wit/world.wit
lxfontes: There is an interaction with `wstd` here and wondering if pebkac or if we have a bigger problem:

https://github.com/bytecodealliance/wstd/blob/main/macro/src/lib.rs#L123
It will always use its internal bindings / world.

The http-api world ends up being the combination of wstd's http + wasmcloud:messaging via this import.
I agree this doesn't give a good representation of the component's world in WIT, but not sure if we can export 2 structs solve it locally without touching wstd.

ricochet: The longterm fix is to essentially make cargo wasi aware. Our options today:

- We have our examples move in lock-step with the wasi version of wstd (rev the component Cargo.toml of dep on wstd, expect to rev wasi version in the wit)
- Have a comment in the world that we're getting wasi-http from wstd

lxfontes: Added the comment in the wit world 👍

ricochet (approved):

PR #283, opened 2026-02-05
feat: TCP Service template

lxfontes:
Similar to #279 , but using TCP between component & service within a workload.

Screenshot 2026-02-05 at 3 11 18 PM


templates/service-tcp/http-api/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+        .json()
+        .await
+        .context("failed to parse body")?;
+
+    let body = task_request.payload.into_bytes();
+
+    let client = wstd::net::TcpStream::connect("127.0.0.1:7777").await?;
+    // write the payload followed by a newline
+    // read the response until a newline is encountered
+    // with a timeout of 5 seconds
```
ricochet: I don't see a timeout?

templates/service-tcp/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[workspace]
+resolver = "3"
+package.edition = "2024"
+package.authors = ["The wasmCloud Team"]
+members = ["service-leet", "http-api"]
+
+[workspace.dependencies]
+wstd = "0.6.3"
```
ricochet:
```suggestion
wstd = "0.6"
```

templates/service-tcp/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+serde = { version = "1.0", features = ["derive"] }
+anyhow = "1.0"
+async-trait = "0.1"
+uuid = { version = "1", features = ["v4"] }
```
ricochet: delete these

templates/service-tcp/http-api/src/lib.rs
```diff
@@ -0,0 +1,75 @@
...
+async fn handle_task(mut req: Request<Body>) -> anyhow::Result<Response<Body>> {
+    let task_request: TaskRequest = req
+        .body_mut()
+        .json()
+        .await
+        .context("failed to parse body")?;
+
+    let body = task_request.payload.into_bytes();
+
+    let client = wstd::net::TcpStream::connect("127.0.0.1:7777").await?;
```
ricochet:
```suggestion
    let client = wstd::net::TcpStream::connect("127.0.0.1:7777")
        .await
        .context("failed to connect to leet service")?;
```

ricochet (approved):
Some cleanup nits, and we can take a pass on this one's README after we align templates.

Could add a buffer limit so that everyone clonging the template follows this as a best practice, but no big deal at this stage.

templates/service-tcp/http-api/src/lib.rs
lxfontes: added 5 secs timeout

PR #285, opened 2026-02-06
feat: Plugin for `wasmcloud:postgres`

lxfontes:
Porting https://github.com/wasmCloud/wasmCloud/tree/main/crates/provider-sqldb-postgres to v2

We can't use v1's example as it performs a direct call into the component via `wash call`.

Validated manually with a combination of existing postgresql example + http hello world.
Screenshot 2026-02-09 at 2 52 19 PM
Screenshot 2026-02-09 at 2 51 40 PM

### DIFF
--- a/crates/wash-runtime/Cargo.toml
+++ b/crates/wash-runtime/Cargo.toml
@@ -15,7 +15,7 @@ path = "src/lib.rs"
 workspace = true
 
 [features]
-default = ["wasi-config", "wasi-logging", "wasi-blobstore", "wasi-keyvalue", "washlet"]
+default = ["wasi-config", "wasi-logging", "wasi-blobstore", "wasi-keyvalue", "wasmcloud-postgres", "washlet"]
 oci = ["dep:oci-client", "dep:oci-wasm", "dep:docker_credential", "dep:sha2", "dep:wit-component"]
 washlet = ["oci"]
 wasi-config = []
@@ -23,6 +23,13 @@ wasi-logging = []
 wasi-blobstore = []
 wasi-keyvalue = []
 wasi-webgpu = ["dep:wasi-webgpu-wasmtime", "dep:wasi-graphics-context-wasmtime"]
+wasmcloud-postgres = [
+    "dep:tokio-postgres", "dep:tokio-postgres-rustls",
+    "dep:deadpool-postgres", "dep:postgres-types",
+    "dep:bigdecimal", "dep:pg_bigdecimal",
+    "dep:bit-vec", "dep:cidr", "dep:geo-types",
+    "dep:ulid", "dep:url", "dep:webpki-roots",
+]
 
 [dependencies]
 anyhow = { workspace = true }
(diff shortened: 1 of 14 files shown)


crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+        )?;
+        bindings::wasmcloud::postgres::prepared::add_to_linker::<_, SharedCtx>(
+            linker,
+            extract_active_ctx,
+        )?;
+
+        Ok(())
+    }
+
+    async fn on_workload_unbind(
```
ricochet, line 518: When a workload is unbound, prepared statements are removed by database name. If two workloads share the same database, unbinding one will delete the other's prepared statements. 

If we add  add component_id to PreparedEntry, store it during prepare, and filter by component IDs (not database names) during cleanup, then two workloads can interact with the same db (expecting that to be the norm).

Another issue, this never removes idle pools from self.pools. Pools (and their connections) persist indefinitely. We should add explicit pool removal when no components reference a database. Potentially we do this as a reaper task that checks async after workload unbind incase another worload is immediately scheduled after being unloaded.

crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+            .iter()
+            .find(|i| i.namespace == "wasmcloud" && i.package == "postgres")
+            .and_then(|i| i.config.get("database").cloned());
+
+        let database = match database {
+            Some(db) => db,
+            None => {
+                bail!("wasmcloud:postgres requires a 'database' config parameter");
+            }
+        };
```
ricochet: minor simplification and one less iterate through interfaces:
```suggestion
        let database = match interfaces
            .iter()
            .find(|i| i.namespace == "wasmcloud" && i.package == "postgres")
        {
            Some(i) => match i.config.get("database").cloned() {
                Some(db) => db,
                None => bail!("wasmcloud:postgres requires a 'database' config parameter"),
            },
            None => return Ok(()),
        };

```

crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+        self.component_databases
+            .read()
+            .await
+            .get(component_id)
+            .cloned()
+    }
+}
+
+/// Extract a query parameter value from a URL string.
+fn extract_query_param<'a>(url: &'a str, key: &str) -> Option<&'a str> {
```
ricochet: we already take on the url crate as a dep. URL encoded values might get a little wonky to handle like for sslmode

crates/wash-runtime/src/plugin/wasmcloud_postgres/mod.rs
lxfontes: moved cleanup to a separate task & having component_id in each preparedstatement

crates/wash-runtime/Cargo.toml
```diff
@@ -81,6 +88,20 @@ oci-wasm = { workspace = true, optional = true, features = ["rustls-tls"] }
...
 
+# PostgreSQL dependencies (optional, behind 'wasmcloud-postgres' feature)
+bigdecimal = { optional = true, workspace = true }
+bit-vec = { optional = true, workspace = true }
+cidr = { optional = true, workspace = true, features = ["std"] }
+deadpool-postgres = { optional = true, workspace = true, features = ["rt_tokio_1"] }
+geo-types = { optional = true, workspace = true }
+pg_bigdecimal = { optional = true, workspace = true }
+postgres-types = { optional = true, workspace = true, features = ["derive", "with-serde_json-1", "with-chrono-0_4", "with-uuid-1", "with-geo-types-0_7", "with-bit-vec-0_6", "with-cidr-0_2", "array-impls"] }
+tokio-postgres = { optional = true, workspace = true, features = [ "runtime", "with-serde_json-1", "with-chrono-0_4", "with-uuid-0_8", "with-geo-types-0_7", "array-impls", "with-bit-vec-0_6", "with-uuid-1" ]  }
```
ricochet: Unused uuid-0_8
```suggestion
tokio-postgres = { optional = true, workspace = true, features = [ "runtime", "with-serde_json-1", "with-chrono-0_4", "with-geo-types-0_7", "array-impls", "with-bit-vec-0_6", "with-uuid-1" ] }
```

ricochet (approved):

PR #288, opened 2026-02-10
chore: moving `initialize_observability` around

lxfontes:
As follow up of https://github.com/wasmCloud/wash/pull/278#issuecomment-3856466039 , moving `initialize_observability` under `wash_runtime::observability`


ricochet (approved):
CARGO_PKG_NAME and CARGO_PKG_VERSION now resolve to: "wash-runtime" / "0.2.0"

Before (in wash crate): resolves to "wash" / "2.0.0-rc.7" (workspace version)

This is probably fine but worth noting.

PR #115, opened 2025-11-05
feat(host): Introduce outbound transport system with initial gRPC client support

Aditya1404Sal:
## Feature or Problem

This PR adds **~~modular~~ transport protocol support** to the HTTP server plugin, with gRPC client support as the first implementation. The new architecture allows WebAssembly components to make outgoing requests using different transport protocols (HTTP/1.1, HTTP/2, gRPC, etc.) through a pluggable handler system.

### Key Changes:

1. **New `OutgoingHandler` trait** - Extensible interface for adding transport protocols
2. **`CompositeOutgoingHandler`** - Chain-of-responsibility pattern for automatic protocol routing
3. **gRPC client support** - HTTP/2-based gRPC transport with TLS and h2c (cleartext) support
4. **Clean separation of concerns** - Transport configuration lives with `HttpServer`, not `HostBuilder`

Components can now make gRPC calls using the [wasmcloud-grpc-client helper crate](https://github.com/Aditya1404Sal/wasmcloud-grpc-client), and the runtime automatically routes requests to the appropriate transport based on headers and URI schemes.

## Related Issues

Closes wasmCloud/wasmCloud#4949

## Release Information

Target: `next` release

### Usage Example:

```rust
// Basic HTTP server (http support included in default behavior, no changes needed)
let http_server = HttpServer::new(router, addr);

let host = HostBuilder::new()
    .with_http_handler(Arc::new(http_server))
    .build()?;
```

### Handler Chain:
```
Outgoing Request → CompositeOutgoingHandler
                   ├─ GrpcHandler (checks content-type: application/grpc)
                   ├─ [Future: WebSocketHandler, custom protocols, etc.]
                   └─ DefaultHttpHandler (fallback)
```

### No Breaking Changes:
- All changes are additive - existing code continues to work
- Existing http/1.1 requests are routed normally
- No changes required for components that don't use gRPC

### Future Extensibility:
This architecture makes it trivial to add new transport protocols without modifying core structures:

```rust
// Example: Adding WebSocket support in the future
let outgoing = CompositeOutgoingHandler::new()
    .with_grpc(grpc_config)?
    .with_handler(Arc::new(WebSocketHandler::new(ws_config)));
```

### Integration Tests
Added three comprehensive integration tests using a test gRPC server and example component (`grpc_hello_world.wasm`):

1. **`test_grpc_client_basic`**
   - Starts test gRPC server (Tonic-based Greeter service)
   - Starts host with gRPC client support
[...]

PR #292, opened 2026-02-19
chore: Refactoring custom wasi implementation

lxfontes:
Making so only `wasi:sockets` is customized.
This allows using the upstream `wasmtime_wasi` crate while customizing socket behaviour.

The only custom code retained is the socket/loopback networking implementation, which was moved into crates/wash-runtime/src/sockets/.

- Deleted crates/wasi/ — removed the entire forked crate
- Created wash-runtime/src/sockets/ — moved custom socket code (TCP, UDP, loopback, address checking) here with granular host trait implementations on `ActiveCtx`
- Updated engine/context — replaced single wasi linker setup with per-interface binding


lxfontes:
This should help #115 ship

crates/wash-runtime/src/engine/mod.rs
ricochet: Hey noticed this is incorrect, namespace is wasi, package is cli

ricochet (approved):
LGTM main thing we'll need to watch out for when doing wasmtime upgrades is that if new interfaces are added, we'll need to include those in our add_wasi_to_linker

PR #293, opened 2026-02-23
Add redis keyvalue plugin support

jfleitz:
### Summary

- Implements a Redis-backed `wasi:keyvalue` plugin (`RedisKeyValue`) in `wash-runtime` using the `redis` crate with a multiplexed async connection. Keys are namespaced per bucket as `{identifier}:{key}`.
- Adds `wasi_keyvalue_redis_url` to `DevConfig` so projects can opt into the Redis backend via `.wash/config.yaml`.
- Wires the Redis backend into `wash dev` with priority order: Redis → filesystem → in-memory.
- Adds a new `examples/redis-keyvalue` project — an HTTP component that stores and retrieves key-value pairs through the `wasi:keyvalue/store` interface, backed by Redis at runtime.

### Changes

| Area | What changed |
|---|---|
| `crates/wash-runtime` | New `redis.rs` plugin; `RedisKeyValue::from_url` constructor; `redis` workspace dep |
| `crates/wash/src/config.rs` | `DevConfig.wasi_keyvalue_redis_url: Option<String>` |
| `crates/wash/src/cli/dev.rs` | Redis branch in keyvalue plugin selection |
| `examples/redis-keyvalue/` | New self-contained example component with WIT deps, config, and README |

### Testing

Testing was completed by running a Redis server locally via Docker:

```bash
docker run --name redis -p 6379:6379 redis:latest
```

The example was built and run with `wash dev` from the `examples/redis-keyvalue/` directory. The following `curl` commands were used to validate correct behaviour:

```bash
# Store a key-value pair
curl -X POST http://localhost:8000 \
  -H "Content-Type: application/json" \
  -d '{"key":"mykey","value":"myvalue"}'
# → 200 Stored key 'mykey'

# Retrieve the value
curl "http://localhost:8000?key=mykey"
# → 200 myvalue

# Retrieve a key that does not exist
curl "http://localhost:8000?key=missing"
# → 404 Key 'missing' not found
```

Signed-off-by: Jeremy Fleitz <jeremy@cosmonic.com>


crates/wash-runtime/src/plugin/wasi_keyvalue/redis.rs
```diff
@@ -0,0 +1,513 @@
...
+            )));
+        };
+        plugin.record_operation("get");
+
+        let bucket_handle = self.table.get(&bucket)?;
+        let mut conn = bucket_handle.conn.clone();
+        let redis_key = bucket_handle.prefixed_key(&key);
+
+        match conn.get::<_, Option<Vec<u8>>>(redis_key).await {
+            Ok(value) => Ok(Ok(value)),
```
lxfontes, line 143: not a blocker: What happens if the key doesn't exist?
as in: the command was successful but there is no data for the key.

crates/wash-runtime/src/plugin/wasi_keyvalue/redis.rs
```diff
@@ -0,0 +1,513 @@
...
+            return Ok(Err(StoreError::Other(
+                "keyvalue plugin not available".to_string(),
+            )));
+        };
+        plugin.record_operation("list_keys");
+
+        let bucket_handle = self.table.get(&bucket)?;
+        let mut conn = bucket_handle.conn.clone();
+        let prefix = bucket_handle.prefix.clone();
+        let pattern = format!("{}:*", prefix);
```
lxfontes: ⚖️  this is analogous to `bucket_handle.prefixed_key("*")`

this is OK given we do have to manipulate prefixes a bit more closely in list_keys.

crates/wash-runtime/src/plugin/wasi_keyvalue/redis.rs
```diff
@@ -0,0 +1,513 @@
...
+                "keyvalue plugin not available".to_string(),
+            )));
+        };
+        plugin.record_operation("increment");
+
+        let bucket_handle = self.table.get(&bucket)?;
+        let mut conn = bucket_handle.conn.clone();
+        let redis_key = bucket_handle.prefixed_key(&key);
+
+        match conn.incr::<_, _, i64>(redis_key, delta as i64).await {
```
lxfontes: can we have u64 instead of i64 for `conn.incr`?
this might truncate `delta` given it's coming as unsigned then casted to signed.

crates/wash-runtime/src/plugin/wasi_keyvalue/redis.rs
```diff
@@ -0,0 +1,513 @@
...
+        plugin.record_operation("get_many");
+
+        let bucket_handle = self.table.get(&bucket)?;
+
+        let values = futures::stream::FuturesOrdered::from_iter(keys.iter().map(|key| {
+            let mut conn = bucket_handle.conn.clone();
+            let redis_key = bucket_handle.prefixed_key(key);
+            let key = key.clone();
+            async move {
+                match conn.get::<_, Option<Vec<u8>>>(redis_key).await {
```
lxfontes: Logic wise this is fine, however can we use [MGET](https://redis.io/docs/latest/commands/MGET/)?

One big redis roundtrip vs many small ones.


The code we have for v1 is for wrpc-bindgen so slightly different signatures https://github.com/wasmCloud/wasmCloud/blob/main/crates/provider-keyvalue-redis/src/lib.rs#L465

crates/wash-runtime/src/plugin/wasi_keyvalue/redis.rs
```diff
@@ -0,0 +1,513 @@
...
+
+        let bucket_handle = self.table.get(&bucket)?;
+
+        let values = futures::stream::FuturesOrdered::from_iter(key_values.iter().map(
+            |(key, value)| {
+                let mut conn = bucket_handle.conn.clone();
+                let redis_key = bucket_handle.prefixed_key(key);
+                let value = value.clone();
+                async move {
+                    match conn.set::<_, _, ()>(redis_key, value).await {
```
lxfontes: similar to get_many (MGET), there is a MSET

examples/redis-keyvalue/.wash/config.yaml
```diff
@@ -0,0 +1,6 @@
+build:
+  command: cargo build --target wasm32-wasip2 --release
+  component_path: target/wasm32-wasip2/release/redis_keyvalue.wasm
+
+dev:
+  wasi_keyvalue_redis_url: redis://127.0.0.1:6379
```
lxfontes: ⚖️ no strong preference, but wondering if we have a full example for redis kv or if we have a single keyvalue example and comments in wash config for different backends.

lxfontes:
This is 99% ✅ 🎉 
The only real blocker is the loss of precision u64 -> i64

crates/wash-runtime/src/plugin/wasi_keyvalue/redis.rs
jfleitz: It will return an Ok(none) from the plugin. Like in the example project, you can then look for `Ok(none)` to handle the case when the key doesn't exist, but the call was successful
[Example redis key-value call](https://github.com/jfleitz/wash/blob/66d97aff41ebdca3138045714c40046a12c27d64/examples/redis-keyvalue/src/lib.rs#L81-L84)

crates/wash-runtime/src/plugin/wasi_keyvalue/redis.rs
jfleitz: Great catch! I checked Redis' docs for the [INCBY command](https://redis.io/docs/latest/commands/incrby/) and Redis actually does used signed integers. I am going to add a validation though to ensure that a large u64 number passed in would throw an error if it would silently wrap/truncate though. That way Redis doesn't mistakenly take it as a large negative number.

jfleitz:
Made updates to use MGET and MSET. I also added a check for u64 numbers that would be truncated/wrapped so that it still meets the WIT spec, but would throw an error if too large. 

Retested with the example/redis-keyvalue project.

lxfontes (approved):

jfleitz:
Changes after approval is to make the key-value example handle all key-value backends in one example Wasm project for https://github.com/wasmCloud/wash/pull/293#discussion_r2850297460

lxfontes (approved):

lxfontes (approved):

